PR #1, opened 2020-11-30
Log unhandled exceptions

swift-kim:
- Log any unhandled exception raised by the .NET runtime. This is useful when `System.DllNotFoundException` is thrown due to unresolved dependency or missing `.so` files.
- Use `InternalLog` instead of `Log` for better debugging experience on TV
- Bump the nuget package version to 1.0.3


xiaowei-guan (approved):
Looks good to me

bbrto21 (approved):
LGTM

PR #6, opened 2020-12-08
Add 'emulators' subcommand to flutter-tizen

HakkyuKim:
This pr adds `flutter-tizen emulators` subcommand feature.

+ `flutter-tizen emulators` : lists created emulators.
+ `flutter-tizen emulators --launch <name>`: launches one of the created emulators with the name `<name>`. `<name>` can also be specified as a prefix of one of the emulators.
+ `flutter-tizen emulators --create --name <name>`: creates a new TV or wearable emulator. If both images are installed, TV profiles have priority over wearable ones. If multiple platform versions of the same profile are installed, the one with the highest version will be chosen.

Limitation

+ `flutter-tizen emulators -h` help message currently describes about creating Android emulators.
+ Android Studio(AVD Manager) must be installed for the command to work.

Both limitations can be resolved by creating the `TizenEmulatorCommand` class that inherits the `FlutterCommand`. This should be revisited when its necessary.

### DIFF
--- a/lib/tizen_emulator.dart
+++ b/lib/tizen_emulator.dart
@@ -75,16 +75,14 @@ class TizenEmulatorManager extends EmulatorManager {
 
     final PlatformImage platformImage = _getPreferredPlatformImage();
     if (platformImage == null) {
-      return CreateEmulatorResult(name,
-          success: false,
-          error:
-              'No suitable Tizen platform images are available. You may need to install these'
-              ' using the Tizen Package Manager.'
-              'Visit the link below to see how to create Tizen platform images for emulators:'
-              ' https://docs.tizen.org/application/tizen-studio/setup/update-sdk/#installing-additional-packages');
+      return CreateEmulatorResult(
+        name,
+        success: false,
+        error: 'No suitable Tizen platform images are available.\n'
+               'You may need to install these using the Tizen Package Manager.',
+      );
     }
 
-    // throwOnError: when should we use or not use?
     final RunResult runResult = await _processUtils.run(
       <String>[
         _tizenSdk.emCli.path,
@@ -238,10 +236,7 @@ class TizenEmulators extends EmulatorDiscovery {
   bool get canListAnything => _tizenSdk?.emCli?.existsSync() ?? false;
 
   @override
-  Future<List<Emulator>> get emulators => _getEmulators();
-
-  // set future and async but not using await?
-  Future<List<Emulator>> _getEmulators() async {
+  Future<List<Emulator>> get emulators async {
     if (!canListAnything) {
       // Although this method doesn't run the `tizenSdk.emCli` command
       // to get emulator info, logically, emulator info shouldn't be
@@ -332,10 +327,13 @@ class TizenEmulator extends Emulator {
   /// See [AndroidEmulator.launch] in [android_emulator.dart] (simplified)
   @override
   Future<void> launch() async {
-    // Null checking on _tizenSdk.emCli.path is already
-    // done at `TizenEmulators._getEmulators`.
+    final String emCliPath = _tizenSdk?.emCli?.path;
+    if (emCliPath == null) {
+      throwToolExit('Unable to locate Tizen Emulator Manager.');
+    }
+
     final List<String> args = <String>[
-      _tizenSdk.emCli.path,
+      emCliPath,
       'launch',
       '--name',
       id,
@@ -343,7 +341,7 @@ class TizenEmulator extends Emulator {
 
     final RunResult runResult = await _processUtils.run(args);
     if (runResult.exitCode == 0) {
-      globals.printStatus('Successfully launched a Tizen emulator, $id.');
+      globals.printStatus('Successfully launched Tizen emulator, $id.');
     } else if (runResult.exitCode == 2) {
       globals.printStatus('Tizen emulator $id is already running.');
     } else {
--- a/lib/tizen_emulator.dart
+++ b/lib/tizen_emulator.dart
@@ -3,13 +3,19 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:async';
+import 'dart:io';
+
 import 'package:file/file.dart';
 import 'package:flutter_tools/src/android/android_sdk.dart';
 import 'package:flutter_tools/src/android/android_workflow.dart';
+import 'package:flutter_tools/src/base/common.dart';
+import 'package:flutter_tools/src/base/logger.dart';
 import 'package:flutter_tools/src/base/process.dart';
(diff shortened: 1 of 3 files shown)


lib/tizen_emulator.dart
```diff
@@ -0,0 +1,346 @@
...
+  Future<List<Emulator>> get emulators => _getEmulators();
+
+  Future<Directory> _getTizenSdkDataDirectory() async {
+    final File sdkInfo = _tizenSdk.directory.childFile('sdk.info');
+    if (await sdkInfo.exists()) {
+      final List<String> lines = await sdkInfo.readAsLines();
+      for (final String line in lines) {
+        final List<String> tokens = line.split('=');
+        if (tokens[0].trim() == 'TIZEN_SDK_DATA_PATH') {
+          return _fileSystem.directory(tokens[1].trim());
```
swift-kim, line 231: I don't really mind but in very rare case the path may contain the "=" letter.

swift-kim (commented):
Nice job! Only minor issues.

A side note:
Historically `flutter-tizen` was designed to extend the existing `flutter` command so this has been perfectly legal:
```
flutter-tizen build apk
flutter-tizen run -d web
```
but apparently the `flutter-tizen emulators` command doesn't stand with this concept any more (you can't use Android emulators with `flutter-tizen`). I'll update your code later if any good solution is found.

HakkyuKim:
@swift-kim 
Thanks for the thoughtful review. I've updated the code based on the suggested improvements. A summary of the changes is listed below:

+ Add ','s for better formatting.
+ Change to sync APIs for simple file operations; `existsSync`, `listSync`, `readAsLinesSync`.
+ Add `sdkDataDirectory` getter in `TizenSdk` class.
+ Move `parseIniLines` function to `tizen_sdk.dart` file and reuse the function to parse `sdk.info` file.
+ The `parseIniLines` function now splits at the first occurrence of the equal sign(=).
+ The `canLaunchAnything` and `canListAnything` getters now evaluate the bool value by checking the existence of the `em-cli` executable.
+ Use `canLaunchAnything` and `canListAnything` getters in null-checking.
+ Handle exitcode from the process that launched the emulator.

Please let me know if I've misunderstood any part of the review, or if further updates are needed.

Commenting on the side note:
I agree that `flutter-tizen emulators` command should eventually work on handling Android emulators. Currently the command can list and launch Android emulators, but can't create them. This is mostly due to the lack of option/flag settings in the `emulator` command. I'll also share any idea if it pops up.

lib/tizen_emulator.dart
```diff
@@ -62,30 +70,39 @@ class TizenEmulatorManager extends EmulatorManager {
...
+              ' using the Tizen Package Manager.'
+              'Visit the link below to see how to create Tizen platform images for emulators:'
+              ' https://docs.tizen.org/application/tizen-studio/setup/update-sdk/#installing-additional-packages');
     }
-    final RunResult runResult = await _processUtils.run(<String>[
-      getEmCliPath(),
-      'create',
-      '-n',
+
+    // throwOnError: when should we use or not use?
```
swift-kim, line 87: If `throwOnError` is true and the process terminates with non-zero exit code, a process exception will be thrown which can be caught by a try-catch block (otherwise it will abort the tool's execution). Whether you should use `throwOnError` or not depends on how you want to control the execution flow. If you don't want the exception to be triggered, you can manually check if the process exited successfully using the value `result.exitCode`. There are some examples in `tizen_device.dart` and other places.

In this specific case, `throwOnError` should be false (default).

lib/tizen_emulator.dart
```diff
@@ -272,42 +285,70 @@ class TizenEmulators extends EmulatorDiscovery {
...
 
+  /// See [AndroidEmulator.launch] in [android_emulator.dart] (simplified)
   @override
   Future<void> launch() async {
-    // TODO(HakkyuKim): launch is overly simplified, for example, it currently doesn't record the emulator process.
-    /// See [AndroidEmulator.launch()] in [android_emulator.dart]
-    final List<String> args = <String>[getEmCliPath(), 'launch', '--name', id];
-    await _processUtils.run(args);
+    // Null checking on _tizenSdk.emCli.path is already
+    // done at `TizenEmulators._getEmulators`.
```
swift-kim, line 336: It is not very clear that `TizenEmulators._getEmulators` is always invoked before `TizenEmulator.launch` is called.

By the way I think this:
```dart
  @override
  Future<List<Emulator>> get emulators => _getEmulators();

  Future<List<Emulator>> _getEmulators() async {
```
can be just
```dart
  @override
  Future<List<Emulator>> get emulators async {
```

swift-kim (commented):
Thanks!

lib/tizen_emulator.dart
```diff
@@ -62,30 +70,39 @@ class TizenEmulatorManager extends EmulatorManager {
...
+    final PlatformImage platformImage = _getPreferredPlatformImage();
     if (platformImage == null) {
       return CreateEmulatorResult(name,
           success: false,
           error:
               'No suitable Tizen platform images are available. You may need to install these'
-              ' using the package manager.');
+              ' using the Tizen Package Manager.'
+              'Visit the link below to see how to create Tizen platform images for emulators:'
+              ' https://docs.tizen.org/application/tizen-studio/setup/update-sdk/#installing-additional-packages');
```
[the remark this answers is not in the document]
HakkyuKim: How about this link? https://docs.tizen.org/application/tizen-studio/common-tools/emulator-manager/#creating-platforms

lib/tizen_emulator.dart
HakkyuKim: `TizenEmulator` class cannot access `TizenEmulators.canLaunchAnything`. I can't think of a good solution at the moment, should I repeat the code like this?

```dart
Future<void> launch() async {
  if(!(_tizenSdk?.emCli?.existsSync() ?? false)){
    throwToolExit('Unable to locate Tizen Emulator Manager.');
  }

  ...
```

lib/tizen_emulator.dart
HakkyuKim: Should have erased the comment there, but I did want to know about the considerations in making the decision, thanks!

lib/tizen_emulator.dart
swift-kim: No need to call `canLaunchAnything`. It's just a null check.

Something like this:
```dart
Future<void> launch() async {
  final String emCliPath = _tizenSdk?.emCli?.path;
  if (emCliPath == null) {
    throwToolExit('Unable to locate Tizen Emulator Manager.');
  }
  final List<String> command = <String>[
      emCliPath,
      ...
```

swift-kim (approved):

PR #8, opened 2020-12-11
Add Dart workflow for continuous integration

swift-kim:


swift-kim:
Quick merging for CI testing.

PR #13, opened 2020-12-21
Misc changes and clean up

swift-kim:
- Make more use of context overrides instead of singletons
- Rename TizenDevices to TizenDeviceDiscovery
- Unsupport Android emulators in TizenEmulatorManager to suppress
  exceptions when Android SDK is missing
- Fix a state error when no platform images are available
- Replace parseIniLines with parseIniFile to avoid a naming conflict
- Make use of TizenWorkflow in tizen_emulator.dart
- General clean-ups

### DIFF
--- a/lib/tizen_doctor.dart
+++ b/lib/tizen_doctor.dart
@@ -4,10 +4,14 @@
 
 import 'package:flutter_tools/src/android/android_studio_validator.dart';
 import 'package:flutter_tools/src/android/android_workflow.dart';
+import 'package:flutter_tools/src/base/context.dart';
 import 'package:flutter_tools/src/doctor.dart';
 import 'package:flutter_tools/src/globals.dart' as globals;
 
 import 'tizen_sdk.dart';
+
+TizenWorkflow get tizenWorkflow => context.get<TizenWorkflow>();
+TizenValidator get tizenValidator => context.get<TizenValidator>();
 
 /// See: [_DefaultDoctorValidatorsProvider] in `doctor.dart`
 class TizenDoctorValidatorsProvider extends DoctorValidatorsProvider {
@@ -19,7 +23,7 @@ class TizenDoctorValidatorsProvider extends DoctorValidatorsProvider {
       // Append before any IDE validators.
       if (validator is AndroidStudioValidator ||
           validator is NoAndroidStudioValidator) {
-        validators.insert(validators.indexOf(validator), TizenValidator());
+        validators.insert(validators.indexOf(validator), tizenValidator);
         break;
       }
     }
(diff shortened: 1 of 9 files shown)


HakkyuKim (approved):
LGTM

lib/tizen_doctor.dart
```diff
@@ -106,11 +110,11 @@ class TizenWorkflow extends Workflow {
...
-  bool get canLaunchDevices => getSdbPath() != null;
+  bool get canLaunchDevices => tizenSdk != null;
 
   @override
-  bool get canListDevices => getSdbPath() != null;
+  bool get canListDevices => tizenSdk != null;
 
   @override
-  bool get canListEmulators => getSdbPath() != null;
+  bool get canListEmulators => tizenSdk != null && tizenSdk.emCli.existsSync();
```
bbrto21, line 119: ```emCli``` seems to need null checking.

lib/tizen_emulator.dart
```diff
@@ -296,65 +292,44 @@ class TizenEmulator extends Emulator {
...
-  final Map<String, String> _properties;
   final ProcessUtils _processUtils;
   final TizenSdk _tizenSdk;
 
-  String _prop(String name) => _properties != null ? _properties[name] : null;
+  @override
+  String get name => id;
+
+  @override
+  String get manufacturer => 'Samsung';
```
bbrto21, line 308: Manufacturer is 'Tizen' on emulator, except on tv-emulator.(I think this is minor things)

lib/tizen_doctor.dart
swift-kim: `emCli` is a null-safe property so this is okay.

lib/tizen_emulator.dart
swift-kim: I only changed the order or the properties. @HakkyuKim Do you have any opinion? It seems the manufacturer value is not available from vm_config.xml. I think it can be just null.

lib/tizen_doctor.dart
bbrto21: Thank you for checking. :+1:

lib/tizen_emulator.dart
HakkyuKim: I originally understood it as the literal meaning of the word manufacturer(the company that made the products). @bbrto21 Could you share any info on where to find this?

Or we can just erase the property, the default value is an empty string.

swift-kim: I think he is referring to the `vendor_name` value from the `sdb capability` result.

HakkyuKim: Indeed the result is "Tizen" for wearable emulators and "Samsung" for TV emulators. Currently I can't find the file that specifies the `vendor_name` value. We can initialize the property value by parsing the output of `sdb capability`, but the emulator has to be launched first. As another option, we can just hard-code the values differently for wearable and TV.

Or if it's not a critical issue, how about revisiting it in the future?

bbrto21: > Or if it's not a critical issue, how about revisiting it in the future?

I agree with you, I think it just minor thing

PR #19, opened 2020-12-29
Fix engine arugments passing on TV devices

swift-kim:
- Rename `TizenBuildInfo.profile` to `TizenBuildInfo.securityProfile` for clarity
- Do not use `__DLP_UNIT_TEST_ARG__` to pass engine arguments to an app any more. Instead, the tool writes the arguments to a plain text file `/home/owner/share/tmp/sdk_tools/{packageId}.rpm` before launching an app and the app reads it on launch (and deletes). This is the only workaround that works for any device type including secure protocol devices. I tried to make the file automatically expire after some time period (10 seconds) but failed after all. No side effect or problem has been found so far.
- Increase the embedding nuget version to 1.1.0
- Add Tizen 4.0 support to the app template
  - Change the API version to 4.0
  - Change the target framework moniker to tizen40
  - Change the C# API version to 4

### DIFF
--- a/lib/commands/build.dart
+++ b/lib/commands/build.dart
@@ -82,7 +82,7 @@ class BuildTpkCommand extends BuildSubCommand with TizenExtension {
     final TizenBuildInfo tizenBuildInfo = TizenBuildInfo(
       buildInfo,
       targetArchs: stringsArg('target-arch'),
-      profile: stringArg('security-profile'),
+      securityProfile: stringArg('security-profile'),
     );
     validateBuild(tizenBuildInfo);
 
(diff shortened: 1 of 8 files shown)


bbrto21:
Does it work on TV emulator too?

swift-kim:
@bbrto21 Sure. I only tested on TV emulator and watches. Needs testing on real TV devices running release images.

bbrto21 (approved):
LGTM

bbrto21:
@swift-kim please resolve conflicts

swift-kim:
Rebased on master.

PR #22, opened 2021-01-05
Add windows&macos platform support

HakkyuKim:
This pr adds support for the Windows and macOS platforms.

- added batch script file to run `flutter-tizen` in cmd without `python3` command
- updated python and dart code to allow running on Windows and macOS.
- updated python and dart code to find files with Windows extension, such as `.exe` and `.bat`

### DIFF
--- a/bin/flutter-tizen
+++ b/bin/flutter-tizen
@@ -4,18 +4,12 @@
 # found in the LICENSE file.
 
 import os
-import platform
 import shutil
 import subprocess
 import sys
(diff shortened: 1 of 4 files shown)


swift-kim (approved):
Additional works to be done:
- Fix the `create` command on Windows
- Update the documentation
- Bump the engine version (to `022b7c35642f8defb20517a3fa2874414983308c` or later)
- Test on macOS

PR #26, opened 2021-01-14
Improve flutter-tizen.bat script

swift-kim:
On some Windows 10 installation, `where python3` returns `C:\Users\[user]\AppData\Local\Microsoft\WindowsApps\python3.exe` which just redirects to the python installation page in Microsoft Store. Therefore we should double check whether `python3 -V` exits with 9009.


HakkyuKim (approved):
LGTM

PR #27, opened 2021-01-19
Add the test command

swift-kim:
- Add the `test` command (for unit/command testing) (tests will be added later)
- Add test-related dependencies in pubspec.yaml
- In bin/flutter-tizen, use `flutter pub` instead of `dart pub` to correctly resolve dependencies of flutter_test
- Remove the use of `waitFor` in `TizenExtension.targetFile` as dart:cli is not available when running tests


HakkyuKim (approved):
LGTM

PR #32, opened 2021-02-02
Define a profile name switch for plugin builds

swift-kim:
- Detect a profile name from a target device and build plugins with a corresponding switch (-DTV_PROFILE or -DWEARABLE_PROFILE) set
- Add a --device-profile option to the "build tpk" command

This patch adds support for plugins like permission_handler which depend on `capi-privacy-privilege-manager` which is not available on TV.

Related to https://github.com/flutter-tizen/plugins/pull/33.

### DIFF
--- a/doc/commands.md
+++ b/doc/commands.md
@@ -46,6 +46,9 @@ Not all commands in the [`flutter`](https://flutter.dev/docs/reference/flutter-c
   # Build a TPK without installing on a device.
   flutter-tizen build tpk
 
+  # Build for a TV device.
+  flutter-tizen build tpk --device-profile tv
+
   # Build for an emulator.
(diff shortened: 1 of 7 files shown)


bbrto21 (approved):
good!

bwikbs:
I like this patch.. but I just wonder if we have consensus on applying profile stuff.
I hope other people also leave comments regarding this.

In my case, agree.

bwikbs (approved):

swift-kim:
Yeah, I made this PR to discuss such stuff.

Another solution we may think of is to build and include two pieces of libxxx_plugin.so for both TV and wearable in a single tpk.

bwikbs:
Now, Are we ready to merge this?

swift-kim:
I think so.

PR #35, opened 2021-02-04
Fix flutter-tizen driver bat file on windows

HakkyuKim:
This pr fixes a bug when `flutter-tizen` runs `python3.bat` instead of `python3.exe`.

On Windows, running a command `program` in cmd can run either `program.exe` or `program.bat`. If both files exist in the same path, `exe` has precedence.

When the `flutter-tizen.bat` script executes on a system and the command `python3` runs `python3.bat`, the control flow will not return back to `flutter-tizen.bat`. To fix this, the `CALL` command should be used. ([MS Docs Reference](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/call))

In summary, it seems that `exe` runs in the current batch process and `bat` runs on a child batch process, but doesn't return to the parent if the `CALL` command is not used.

To handle both extension files, `CALL` command should always be used for calling `python3`.

### DIFF
--- a/bin/flutter-tizen.bat
+++ b/bin/flutter-tizen.bat
@@ -10,13 +10,13 @@ where /q python3 || (
     ECHO Error: Unable to find python3 in your PATH.
     EXIT /B 1
 )
-python3 -V >NUL 2>NUL
-IF errorlevel 9009 (
+
+CALL python3 -V >NUL 2>NUL || (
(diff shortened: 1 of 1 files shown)


swift-kim (approved):
LGTM.

bbrto21 (approved):
LGTM

PR #36, opened 2021-02-10
Remove the use of FlutterRotateWindow

swift-kim:
`FlutterRotateWindow` is deprecated and does nothing in the current version of the engine.


bbrto21 (approved):
LGTM

PR #37, opened 2021-02-10
Handle multiline logs properly

swift-kim:
Currently lines that do not start with `*/ConsoleMessage(*):` are all being skipped although some of them are parts of multiline messages. Use a similar logic to AdbLogReader to handle them properly.


bbrto21 (approved):
LGTM

PR #42, opened 2021-02-19
Use deviceTemplate name as an emulator name

swift-kim:
Before:
```
$ flutter-tizen emulators
5 available emulators:

W-5.5-circle-x86         • W-5.5-circle-x86         • Samsung • linux
T-samsung-6.0-x86        • T-samsung-6.0-x86        • Samsung • linux
M-6.0-x86                • M-6.0-x86                • Samsung • linux
W-4.0-circle-x86-preview • W-4.0-circle-x86-preview • Samsung • linux
W-6.0-circle-x86         • W-6.0-circle-x86         • Samsung • linux
```

After:
```
$ flutter-tizen emulators
5 available emulators:

W-5.5-circle-x86         • Wearable Circle • Samsung • linux
T-samsung-6.0-x86        • HD1080 TV       • Samsung • linux
M-6.0-x86                • HD Mobile       • Samsung • linux
W-4.0-circle-x86-preview • Wearable Circle • Samsung • linux
W-6.0-circle-x86         • Wearable Circle • Samsung • linux
```

Also fixes a misleading output message when an emulator is not already running but fails to launch (`exitCode == 2`) for some reason.

### DIFF
--- a/lib/tizen_emulator.dart
+++ b/lib/tizen_emulator.dart
@@ -249,18 +249,22 @@ class TizenEmulators extends EmulatorDiscovery {
     TizenEmulator loadEmulatorInfo(String id) {
       final File configFile =
           emulatorDir.childDirectory(id).childFile('vm_config.xml');
-      final Map<String, String> properties = <String, String>{};
 
       final XmlDocument xmlDocument =
           XmlDocument.parse(configFile.readAsStringSync());
-      final String name = xmlDocument.findAllElements('name').first.text;
+      final XmlElement deviceTemplate =
+          xmlDocument.findAllElements('deviceTemplate').first;
+      final String name = deviceTemplate.getAttribute('name');
       final XmlElement diskImage =
           xmlDocument.findAllElements('diskImage').first;
       final String profile = diskImage.getAttribute('profile');
       final String version = diskImage.getAttribute('version');
-      properties['name'] = name;
-      properties['profile'] = profile;
-      properties['version'] = version;
+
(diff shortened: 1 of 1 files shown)


lib/tizen_emulator.dart
```diff
@@ -326,10 +334,11 @@ class TizenEmulator extends Emulator {
         await _processUtils.run(<String>[emCli.path, 'launch', '--name', id]);
```
HakkyuKim, line 334: > Also fixes a misleading output message when an emulator is not already running but fails to launch (exitCode == 2) for some reason.

I think this is because the emulator's CPU-VT option (Emulator Manager > Edit > HW Support) is set to OFF. (It seems that some emulators are created with OFF by default)

we can switch this option to ON with the following command
```
em-cli modify -n <emulator name> -w yes
```
How about adding something like the following before calling `emCli launch`?
```dart
// emulator's CPU-VT property must be set to ON to launch 
await _processUtils.run(<String>[emCli.path, 'modify', '--name', id, '-w', 'yes']);
```

Though, it is strange because the help message of `em-cli modify` says that the default option for -w is "yes". I  guess it is not properly applied.

HakkyuKim (approved):
LGTM

lib/tizen_emulator.dart
HakkyuKim: Resolved. There seems to be several issues that generate exit code 2. This would need to be investigated another time. The current code is sufficient.

bbrto21 (approved):
LGTM

PR #47, opened 2021-02-22
Reduce build time by using cache files

HakkyuKim:
This pr leverages the caching functionality of the `FlutterBuildSystem`
to reduce the overall build time by re-building only the modified files.

## Summary

The `FlutterBuildSystem` supports caching intermediate build files
of each target to reduce the overall build time.
In order to use this functionality, we need to specify the files
that each target depends on. If those files are always known during the
dev-lifecycle, they can be listed in `Target.inputs` or `Target.outputs`.
In most cases, however, new dependencies such as plugins or source
code files will be added during development. To handle these cases,
we can either (1) use wildcard patterns in `Source.pattern()`, or (2)
create a `DepFile` that stores the dependencies of a target at runtime
and specify the `DepFile` filename in `Target.depfiles`.

Although (1) is a simpler solution, the implementation of resolving
the wildcard pattern doesn't handle our use cases(See `visitPattern()`
in `source.dart`), such as:
- doesn't support recursive file search
- throws error when intermediate directories are missing in the file path

We need at least one of the above functionalities because our intermediate
files are structured in a multiple depth folder hierarchy.

As such, I've chosen to use solution (2) despite increasing the code
complexity of maintaining depfiles. It is crucial for depFiles to list
all input and output files that participated in `Target.build()`.

## Correctness Check

The following checks are performed for `DotnetTpk` (Debug/Release). 
(`NativeTpk` is left for future work.)
- [x] Skipping all build steps if nothing has changed
- [x] Rebuilding only necessary files when app source changes
- Rebuilding only necessary files when adding or removing plugins
    - [x] Adding new plugin
    - [x] Removing plugin
    - [ ] Plugin version changes
    - [ ] Plugin code changes for path referenced plugins
    - [ ] Plugin not rebuilding when non-plugin packages are added or removed from .packages
- Rebuilding tpk when non-generated files in the host app folder changes
    - [x] Rebuilding when source code changes 
    - [x] Rebuilding tpk when certificate changes
    - [x] Rebuilding tpk when manifest changes

### Before

| Build Mode | First Build | Second Build |
|---|---|---|
|Debug| 18.1s | 13.0s|
|Release|19.0s|5.4s|

### After

| Build Mode | First Build | Second Build |
|---|---|---|
|Debug| 19.1s | 1.4s|
|Release|19.9s|0.6s|

### DIFF
--- a/lib/tizen_build_target.dart
+++ b/lib/tizen_build_target.dart
@@ -10,6 +10,7 @@ import 'package:flutter_tools/src/base/common.dart';
 import 'package:flutter_tools/src/base/process.dart';
 import 'package:flutter_tools/src/build_info.dart';
 import 'package:flutter_tools/src/build_system/build_system.dart';
+import 'package:flutter_tools/src/build_system/depfile.dart';
 import 'package:flutter_tools/src/build_system/exceptions.dart';
 import 'package:flutter_tools/src/build_system/source.dart';
 import 'package:flutter_tools/src/build_system/targets/android.dart';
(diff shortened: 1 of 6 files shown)


lib/tizen_build_target.dart
```diff
@@ -24,6 +25,18 @@ import 'tizen_project.dart';
 import 'tizen_sdk.dart';
 import 'tizen_tpk.dart';
 
+Future<void> writeDepFile(String depFileName, Environment environment,
+    List<File> inputFiles, List<File> outputFiles) async {
```
swift-kim, line 29:
```suggestion
void _writeDepfile(
    String name,
    Environment environment,
    List<File> inputs,
    List<File> outputs,
) {
```

PR #56, opened 2021-03-15
Update for Flutter 2.0.1

swift-kim:
- Update the Flutter SDK and engine hash
  - Engine migration issues: https://github.com/flutter-tizen/engine/issues/49
  - Engine release: https://github.com/flutter-tizen/engine/releases/tag/113f51b
- Various fixes for the updated flutter_tools
  - Do not measure progress timeout
  - Display null safety mode when building an app
  - Change the code size analysis output directory
  - Remove obsolete `ApplicationPackageStore`
  - Add new subcommands in the `pub` command
  - etc.
- Code clean-ups
  - Rename `TizenNativeProject` to `TizenLibrary` for clarity
  - Rename `activeProfile` to `active`
  - Refector `create.dart` based on the refactored `CreateBase`
  - etc.
- Update `devtools-usage.md` since the usage has been changed
- Update template_manifest.json and other miscellaneous files

PR #58, opened 2021-03-17
Fix emulators command crashes

swift-kim:
An exception is thrown when there exists an emulator image whose
profile is not both TV and wearable (e.g. mobile).

This change was originally part of #56, but made into a separate PR for clarity.


HakkyuKim (approved):
LGTM

Issue #57, opened 2021-03-16
Native app host performance issue


swift-kim:
Comparison with the previous result:

| Launching<br>time (ms) | Native | .NET (loader) | .NET<br>(standalone) |
|:-:|:-:|:-:|:-:|
| Jul 2020 | 562 | 577 | - |
| Mar 2021 | **1178** | **849** | 1316 |

| PSS (KB) | Native | .NET (loader) | .NET<br>(standalone) |
|:-:|:-:|:-:|:-:|
| Jul 2020 | 20183 | 38336 | - |
| Mar 2021 | 29509 | 46837 | 45634 |

(measured with the same jsonexample app on TW3)

The regression over time is somewhat understandable because our embedder implementation has grown a lot during the period. However, according the recent measurements, native host apps were even slower than .NET host apps unexpectedly. Need to investigate why.

Tip. You may also use this command to trace detailed start-up time events:
```
flutter-tizen run --profile --trace-startup
```

swift-kim:
This specific issue has been fixed by #62. We'll look into other performance-related issues later on with more detailed measurements.

PR #62, opened 2021-03-23
add linker option "-Wl,-E" to remove regression

wscho77:
When the toolchain is changed, the linker option "-Wl,-E" is omitted.
Due to this, performance degradation occurred by launching the app using exec() in the candidate process.

Add "-Wl,-E" to linker option of cpp template so that main() symbol are added to the dynamic symbol table,
and can be used in candidate process.

| Applying  "-Wl,-E" linker option   |   Startup Time  |
|:-:|:-:|
| X                                 |  1162 ms              |
| O |  811 ms |

#57

### DIFF
--- a/templates/app/cpp/project_def.prop
+++ b/templates/app/cpp/project_def.prop
@@ -17,7 +17,7 @@ USER_CPP_UNDEFS =
 # Compiler/linker flags
 USER_CFLAGS_MISC =
 USER_CPPFLAGS_MISC = -c -fmessage-length=0
-USER_LFLAGS =
+USER_LFLAGS = -Wl,-E
 
 # Libraries and objects
 USER_LIB_DIRS = lib


wscho77:
According to my additional test, adding attribute **hw-acceleration="on"** to the **ui-application** element of tizen-manifest.xml file can improve startup performance (about 100ms).

The hw-acceleration configuration seems to call the **elm_config_accel_preference_set()** function internally.
But, I don't know how EFL related setting can makes the flutter app faster.

swift-kim:
Great!

Is `main()` directly called by the loader (candidate) process without fork-exec if its symbol can be found in the dynamic symbol table?

wscho77:
> Great!
> 
> Is `main()` directly called by the loader (candidate) process without fork-exec if its symbol can be found in the dynamic symbol table?

Yes. According to the logic of the candidate process, if main() can be found by dlsym(), then main() function is called directly. Otherwise, the executable is executed using execv().

swift-kim (approved):

swift-kim:
Thanks for the patch!

PR #65, opened 2021-03-30
Allow plugins to define both pluginClass and dartPluginClass

swift-kim:
Hybrid plugins may need some extra initialization in Dart code
(e.g. for federated plugins to instantiate platform interfaces).

This change may be required by the notifications plugin currently being investigated by @nojinjeong.


swift-kim:
Now the `fileName:` attribute is not required for Dart plugins.

The tool will always find the static `{{PluginClass}}.register()` method from `package:{{project_name}}/{{project_name}}.dart` file but not `package:{{project_name}}/{{fileName}}`.

HakkyuKim (approved):
LGTM

PR #69, opened 2021-04-06
Modernize the C++ app template

swift-kim:
The C++ template will look similar to the C# template when done.


swift-kim:
Implementing a simple client wrapper for Tizen is done, which only contains the `FlutterApp` class for now.

Any existing native project should still work unless it has Tizen plugin dependencies. It is recommended to re-generate the project by removing `tizen` directory and running `flutter-tizen create` though.

Native apps cannot target both Tizen 4.0 and Tizen 5.5+ at the same time so only Tizen 5.5+ is support by the native template right now. (c.f. .NET apps choose the right embedder for the current platform version and load it dynamically at runtime.)

embedding/cpp/flutter_app.cc
```diff
@@ -0,0 +1,198 @@
...
+    return FlutterDesktopGetPluginRegistrar(handle, plugin_name.c_str());
+  }
+  return nullptr;
+}
+
+void FlutterApp::ParseEngineArgs() {
+  char *app_id;
+  if (app_get_id(&app_id) != 0) {
+    return;
+  }
```
pkosko, line 181: Maybe adding error log will be valuable for debugging this case? (I didn't compile below suggestion, just written it here by hand).
```suggestion
  int ret = app_get_id(&app_id);
  if (ret != APP_ERROR_NONE) {
    TizenLog::Error("Error gathering application id: %d (%s)", ret, get_error_message(ret));
    return;
  }
```

pkosko (commented):
Minor suggestion only.

embedding/cpp/flutter_app.cc
swift-kim: Thanks! I'll fix.

embedding/cpp/flutter_app.cc
```diff
@@ -0,0 +1,199 @@
...
+                                   &window_prop.width) != 0 ||
+      system_info_get_platform_int("http://tizen.org/feature/screen.height",
+                                   &window_prop.height) != 0) {
+    TizenLog::Error("Could not obtain the screen size.");
+    return false;
+  }
+
+  std::string res_path;
+  {
+    auto path = app_get_resource_path();
```
bbrto21, line 31: app_get_resource_path can return null pointer. I think it would be better to check the return value.

swift-kim: Thanks!

bbrto21 (approved):
LGTM!

embedding/cpp/flutter_app.cc
bbrto21: @swift-kim is it done?

swift-kim: Done!

PR #73, opened 2021-04-08
Ensure tizen-manifest.xml exists for app project

swift-kim:
Fixes a null reference exception when running the build command in a plugin project directory. The root cause is `_createEntrypoint` not checking the validity of project properly. The correct behavior is printing this error:

```
Target file "lib/main.dart" not found.
```


HakkyuKim (approved):
LGTM

Issue #74, opened 2021-04-09
Permission denied error if an empty lib directory exists in tpk


swift-kim:
![image](https://user-images.githubusercontent.com/40190588/114120477-88090980-9927-11eb-81cb-4e224fb26efb.png)

```sh
$ ls -l /tmp/flutter_tools.OPSIYV/YEBBPE/lib
total 8
d--------- 2 swift swift 4096  4월  9 11:27 arm
drwxrwxr-x 2 swift swift 4096  4월  9 11:28 x86

$ ls -l /tmp/flutter_tools.OPSIYV/YEBBPE/lib/arm/
ls: cannot open directory '/tmp/flutter_tools.OPSIYV/YEBBPE/lib/arm/': Permission denied
```

swift-kim:
This must have been fixed by #76.

PR #76, opened 2021-04-09
Clear tpkroot directory before dotnet build

HakkyuKim:
Solves issue #74.

It seems that dotnet cli or the build script that is run by the cli doesn't remove empty directories when updating `tpkroot`. I cannot think of a way to fix this without changing the behavior of the dotnet build, a quick solution is deleting `tpkroot` before running dotnet build.


swift-kim (approved):
LGTM if it works reliably.

PR #78, opened 2021-04-09
Add support for <service-application> in TizenManifest

wiertel:
This change allows TizenManifest to read Application ID from
<service-application> element as well as <ui-application>.

It is a small step towards supporting [Tizen Service Application](https://docs.tizen.org/application/native/guides/applications/service-app/). It prevents a crash when flutter-tizen installs application which has only `service-application` instead of `ui-application` in `tizen-manifest.xml`.

### DIFF
--- a/lib/tizen_tpk.dart
+++ b/lib/tizen_tpk.dart
@@ -114,7 +114,7 @@ class TizenTpk extends ApplicationPackage {
 ///
 /// See: [ApkManifestData] in `application_package.dart`
 class TizenManifest {
-  TizenManifest(this._document);
+  TizenManifest(this._document) : applicationId = _findApplicationId(_document);
 
   factory TizenManifest.parseFromXml(File xmlFile) {
     if (xmlFile == null || !xmlFile.existsSync()) {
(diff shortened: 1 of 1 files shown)


lib/tizen_tpk.dart
```diff
@@ -158,10 +158,39 @@ class TizenManifest {
...
-    return parent.getAttribute('appid');
+    // storing _applicationId to avoid showing warning more than once
+    if (_applicationId != null) {
+      return _applicationId;
+    }
+    final Iterable<XmlElement> uiappe =
+        _manifest.findElements('ui-application');
+    final Iterable<XmlElement> srvappe =
+        _manifest.findElements('service-application');
+    if (uiappe.isEmpty && srvappe.isEmpty) {
```
wiertel, line 169: actually, if manifest does not contain any of those, the build process fails earlier so this code is not really usefull

swift-kim (commented):
This is just my opinion. Don't take too seriously!

Thanks!

wiertel:
Based on review comments, I've changed the implementation. I don't mind that you want the code to be clean. It's a good thing. Now the logic and possible error/warning reporting is done during the construction.
I'm not sure about member order in the source code. I'm open to suggestion.

swift-kim (approved):

Issue #83, opened 2021-04-12
[arm64] Current status on rpi4


swift-kim:
- [ ] The app takes up only 2/3 of the display.

  ![20150322-125958](https://user-images.githubusercontent.com/40190588/114368308-45119500-9bb8-11eb-9a47-f25281a76b76.jpg)

- [x] Building native tpk fails with the "file format not recognized" error.

  ```sh
  $ flutter-tizen run
  ...
    Building target: Debug/runner
    Invoking: C/C++ Linker
  "/home/swift/tizen-studio/tools/smart-build-interface/../llvm-10/bin/clang++" -o ./Debug/runner @./Debug/target_objs.file   -Llib
  -L"/home/swift/Git/flutter/dummy/arm64_native/tizen/lib" -Wl,--no-undefined  -Wl,-E -target arm-tizen-linux-gnueabi -gcc-toolchain
  "/home/swift/tizen-studio/tools/smart-build-interface/../arm-linux-gnueabi-gcc-9.2/" -ccc-gcc-name arm-linux-gnueabi-g++ -march=armv7-a -mfloat-abi=softfp
  -mfpu=vfpv3-d16 -mtune=cortex-a8 -Wl,-z,relro,--as-needed  -Xlinker --as-needed -pie -lpthread
  --sysroot="/home/swift/tizen-studio/tools/smart-build-interface/../../platforms/tizen-5.5/wearable/rootstraps/wearable-5.5-device.core" -Xlinker
  --version-script="./.exportMap" -L/home/swift/tizen-studio/platforms/tizen-5.5/wearable/rootstraps/wearable-5.5-device.core/usr/lib @./Debug/platform_libs.file
  -Xlinker -rpath='$ORIGIN/../lib' -Werror-implicit-function-declaration -lflutter_tizen -L/home/swift/Git/flutter/dummy/arm64_native/tizen/lib
  -I/home/swift/Git/flutter/flutter-tizen/bin/cache/artifacts/engine/common/client_wrapper/include
  -I/home/swift/Git/flutter/flutter-tizen/bin/cache/artifacts/engine/common/public -I/home/swift/Git/flutter/flutter-tizen/embedding/cpp/include -DCOMMON_PROFILE
  -Wl,-unresolved-symbols=ignore-in-shared-libs
  lib/libflutter_tizen.so: file not recognized: file format not recognized
  clang++: error: linker command failed with exit code 1 (use -v to see invocation)
[...]

PR #84, opened 2021-04-12
Support arm64 architecture

swift-kim:


swift-kim:
The tool side work is pretty much done. Only need to update the engine hash when the new engine is released.

PR → https://github.com/flutter-tizen/engine/pull/63

lib/tizen_build_target.dart
```diff
@@ -273,9 +273,10 @@ class TizenPlugins extends Target {
         sharedLib.copySync(outputDir.childFile(sharedLib.basename).path);
 
         // copy binaries that plugin depends on
-        final Directory pluginLibDir = pluginDir
-            .childDirectory('lib')
-            .childDirectory(arch == 'arm' ? 'armel' : 'i586');
+        final String pluginArch =
+            arch == 'arm' ? 'armel' : (arch == 'x86' ? 'i586' : arch);
+        final Directory pluginLibDir =
+            pluginDir.childDirectory('lib').childDirectory(pluginArch);
```
swift-kim, line 279: @HakkyuKim For Tizen plugins with shared libs, which directory should I place arm64 libs?

- arm: armel
- x86: i586
- arm64: aarch64?

HakkyuKim: aarch64. Tizen CLI uses that name internally.

Tizen CLI uses the "architecture" attribute in the rootstrap xml file to replace {BUILD_ARCH} variable. This may allow us to use the names arm and x86, but the attribute value is used in several places internally so we should leave the names to armel and i586 at the moment.

swift-kim:
@HakkyuKim Could you test this change by running any app either on Tizen 4.0 emulator or Tizen 4.0 device? If the app runs without error, then it's fine.

lib/tizen_sdk.dart
```diff
@@ -112,10 +112,9 @@ class TizenSdk {
     String profile,
     @required String arch,
   }) {
-    final String type = arch == 'x86' ? 'emulator' : 'device';
     final String id = profile == null
-        ? 'wearable-$defaultTargetPlatform-$type.flutter'
-        : '${profile.replaceFirst('common', 'wearable')}-$type.flutter';
+        ? 'wearable-4.0-$arch.flutter'
+        : '${profile.replaceFirst('common', 'wearable')}-$arch.flutter';
```
wiertel, line 117: this substitution makes error regarding missing `rootstraps/*.flutter.xml` harder to understand, at least that was my experience
I'm not suggesting change here, I think I will submit an issue about the error message.

swift-kim: I'm actually aware of the error and you can simply remove the broken links to work around. Sorry.

swift-kim: Actually I couldn't find a very good way to remove those stale files automatically.

HakkyuKim:
> @HakkyuKim Could you test this change by running any app either on Tizen 4.0 emulator or Tizen 4.0 device? If the app runs without error, then it's fine.

I've tested the changes on Tizen 4.0 emulator with a few plugin example apps, I don't have access to a Tizen 4.0 device at the moment. On an emulator, it works fine.

lib/tizen_sdk.dart
swift-kim: I think I was referring to a different error regarding stale rootstraps. The error you are referring to must be generated in these conditions:

- The user tries to build a native plugin (or app) without installing a target platform SDK (e.g. Tizen 4.0 SDK for Tizen 4.0 app). In this case the user can just run `flutter-tizen doctor` and follow its instructions.
- The user tries to build a native plugin (or app) for unknown platform version (such as 6.5) or architecture.

In either case it would be better to properly warn the user of the missing rootstrap instead of printing error messages that are usually hard to understand.

swift-kim:
Any review please?

lib/commands/build.dart
```diff
@@ -31,7 +31,7 @@ class BuildTpkCommand extends BuildSubCommand with TizenExtension {
       'target-arch',
       splitCommas: true,
       defaultsTo: <String>['arm'],
-      allowed: <String>['arm', 'aarch64', 'x86'],
+      allowed: <String>['arm', 'arm64', 'x86'],
```
xuelian-bai, line 34: Just curious, why replace aarch64 with arm64?

swift-kim: Because

- It's more consistent (arm-arm64-x86) and easier to remember for users
- The Flutter engine uses the term (`flutter/tools/gn --linux-cpu arm64 ...`) and so does our CI

HakkyuKim (approved):

lib/commands/build.dart
```diff
@@ -58,9 +58,6 @@ class BuildTpkCommand extends BuildSubCommand with TizenExtension {
 
   /// See: [android.validateBuild] in `build_validation.dart`
   void validateBuild(TizenBuildInfo tizenBuildInfo) {
-    if (tizenBuildInfo.targetArchs.contains('aarch64')) {
```
HakkyuKim, line 61: Found this after approving, maybe we can block aarch64 for native projects?

swift-kim: The current limitation (#83) not only applies to app builds but also to plugin builds, so `build.dart` is not a good place to add the error condition.

swift-kim: I fixed the aarch64 native build. The iot-headed-6.0 rootstrap is required.

lib/tizen_sdk.dart
swift-kim: I improved the error messages when the target rootstrap definition is missing (for both cases in the above).

lib/commands/build.dart
swift-kim: Now we use "arm64" instead of "aarch64" in most cases (unless we're interacting with the Tizen CLI or building a .NET tpk).

swift-kim:
Sorry for spamming the commits but please review the new changes!

HakkyuKim (approved):
arm64 build succeeds for native projects. LGTM!

PR #85, opened 2021-04-15
Move Tizen.Flutter.Embedding into embedding directory

swift-kim:
- Move the directory
- Use `ApplicationId` instead of `PackageId` in `ParseEngineArgs()` for consistency with the native impl ([`FlutterApp`](https://github.com/flutter-tizen/flutter-tizen/blob/master/embedding/cpp/flutter_app.cc))
- Add `IPluginRegistry`
- Resolves the latest Tizen.Flutter.Embedding if `FlutterEmbeddingVersion` is not provided (use floating version to allow restoring with the .NET CLI or Visual Studio)
- Other misc changes for consistency with the native impl (e.g. debug asserts)


swift-kim:
Would it be better to use `ProjectReference` (in `Runner.csproj`) rather than the current `PackageReference` for ease of debugging? There are several options to pass the path to Tizen.Flutter.Embedding to `Runner.csproj` but either of them will increase complexity of the tool code (mainly because Windows doesn't support symlinks in user mode by default).

embedding/csharp/IPluginRegistry.cs
```diff
@@ -0,0 +1,13 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+using System;
+
+namespace Tizen.Flutter.Embedding
+{
+    public interface IPluginRegistry
+    {
+        public IntPtr GetRegistrarForPlugin(string pluginName);
```
bbrto21, line 11: I think it would be better to add the ```abstract``` keyword

```suggestion
        public abstract IntPtr GetRegistrarForPlugin(string pluginName);
```

embedding/csharp/FlutterApplication.cs
```diff
@@ -106,90 +105,89 @@ protected override void OnCreate()
...
 
-        public IntPtr GetPluginRegistrar(string pluginName)
+        /// <summary>
+        /// Returns the plugin registrar handle for the plugin with the given name.
+        /// The name must be unique across the application.
+        /// </summary>
+        public IntPtr GetRegistrarForPlugin(string pluginName)
         {
-            if (!Handle.IsInvalid)
+            if (Handle)
```
bbrto21, line 190: Why didn't you use ```Debug.Assert(Handle)``` here?

embedding/csharp/IPluginRegistry.cs
swift-kim: In C#, as far as I know, all methods in interfaces are abstract by default and it's illegal to use the `abstract` keyword.

bbrto21: Um... abstract keyword in method declaration, I remember that was a keyword that forced implementation like pure virtual function in c++.
I'm sorry if I was wrong :) and If so, don't mind this opinion

bbrto21: >  all methods in interfaces are abstract by default

oh.. I understood

embedding/csharp/FlutterApplication.cs
swift-kim: Conclusion: There's no exact reason.

Some client wrapper implementations for platforms like [glfw](https://github.com/flutter-tizen/engine/blob/31279eae37874150064da5983425c7769f0b0f40/shell/platform/glfw/client_wrapper/flutter_window_controller.cc#L81-L91) and windows explicitly check if the engine is currently running and emit error messages if not. Some platforms like [iOS](https://github.com/flutter-tizen/engine/blob/31279eae37874150064da5983425c7769f0b0f40/shell/platform/darwin/ios/framework/Source/FlutterAppDelegate.mm#L237-L243) check if the instance is null but do not log errors, and some platforms like macOS just do not check anything. (And Tizen is the only platform that uses debug time asserts.) Sometimes things are different when it comes to lifecycle methods instead of plugin registry methods, but in general, the checks are more strict in case of plugin registry methods, maybe because each platform ensures `OnDestroy()` is called only when `OnCreate()` has succeeded.

bbrto21 (approved):
LGTM

Issue #94, opened 2021-04-26
Support service application type


rwalczyna:
The goal is to add support for Tizen Service Application that could run Dart code in the background without UI and communicate with other Flutter applications.

Main capabilities:

* Performing complex / time consuming headless computation in separate process
* Starting service application from different Flutter application
* Sending commands from different Flutter application to the service
* Communication between service and other Flutter applications

TODO:

Changes in tools:

- [ ] [engine](https://github.com/flutter-tizen/engine) - adding support for headless application (IN PROGRESS: flutter-tizen/engine#66)
- [ ] [flutter-tizen](https://github.com/flutter-tizen/flutter-tizen/) - creation and running tizen service application (preparing templates)

Changes in application runner:
- [ ] Tizen service application with service_app_main instead of ui_app_main
- [ ] Starting service application and receiving simple commands (service_app_control_cb)
- [ ] Bi-directional communication using Message Port (probably creation of flutter plugin)

Related native solutions: [Service Application](https://docs.tizen.org/application/native/guides/applications/service-app/), [Message Port](https://docs.tizen.org/application/native/guides/app-management/message-port/), [Application Control](https://docs.tizen.org/application/web/guides/app-management/app-controls/)

swift-kim:
I have a few questions:

1. https://docs.tizen.org/application/native/guides/applications/service-app/#attribute says the `auto-restart` and `on-boot` features are no longer supported on all Tizen devices. Should a service app always be launched manually from the command line? I also read https://docs.tizen.org/application/native/guides/applications/efl-ui-app/#background-categories but it's still unclear. (Should a service app be always packaged with a UI app to operate properly?)
2. https://docs.tizen.org/application/native/guides/applications/service-app/#application-states says a service app can go into the suspended state. Do you know any details about it, like how the suspended app can wake up and resume the execution?

rwalczyna:
1. These two features are not available on wearable - due to battery consumption optimization. On other devices it should work properly. I am testing it on TM1 and it works as expected. It is not documented, but I think that on wearable developer should use workaround to be sure that application runs properly (setting alarm to reopen app or using companion UI application).
Regarding `background-category` it looks like without it application is blocked by system after a while (to save resources).
[...]

PR #96, opened 2021-04-28
Add support for service tizen application

rwalczyna:
It's a part of #94. PR for engine flutter-tizen/engine#66 has to land first.

`--tizen-service` selector for `flutter-tizen create` has been added

It enables to create headless Flutter application in Tizen. Only cpp application type is supported 

Usage:
`flutter-tizen create --tizen-language cpp --tizen-service app_service`

Example main.dart file of service application for testing purposes:

```dart
int counter_ = 0;

Future<void> main() async {
  while (true) {
    print('Counter: $counter_');
    await Future.delayed(Duration(seconds: 1));
    counter_ += 1;
  }
}
```


swift-kim:
It seems you have figured out how the `create` command works by yourself. Good job! The logic is in fact not very straightforward and I'm sorry if anything has confused you. Actually there wasn't any better option when I was implementing it.

Just one thing to discuss:

The type `FlutterApp` is ambiguous. Wouldn't it be better to create an explicit type `FlutterServiceApp` (either from scratch or by extending `FlutterApp`)? Maybe your intention was to share the same files under `templates/app/cpp` across the different app types. However, creating an individual template for each app type (e.g. `templates/ui-app/cpp/...` and `templates/service-app/cpp/...`) looks like a better option for me. You will have to refactor the `create.dart` code a little bit (by replacing the `for` loop with something different) for this.

pwasowski2 (commented):
LGTM 👏

templates/app/cpp/tizen-manifest-service.xml.tmpl
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest package="{{androidIdentifier}}" version="1.0.0" api-version="4.0" xmlns="http://tizen.org/ns/packages">
+    <profile name="common"/>
+    <service-application appid="{{androidIdentifier}}" exec="runner" type="capp" multiple="false" nodisplay="true" taskmanage="false" auto-restart="false">
```
pkosko, line 4: Probably this makes no issue, but in [docs](https://developer.tizen.org/ko/development/tizen-studio/native-tools/configuring-your-app/manifest-text-editor?langredirect=1#service) there is no 'nodisplay' member in case of service-application tag, but probably it will be just ignored.

templates/app/cpp/tizen-manifest-service.xml.tmpl
```diff
@@ -0,0 +1,10 @@
...
+        <label>{{projectName}}</label>
+        <icon>ic_launcher.png</icon>
+        <background-category value="media"/>
```
pkosko, line 7: According to [this](https://developer.tizen.org/ko/development/tizen-studio/web-tools/configuring-your-app/configuration-editor?langredirect=1) besides of adding background-category field there should be also a <tizen:setting background support="enable" /> to make the behaviour correct. But as I am not tizen-manifest expert, thus maybe this is not true for service applications.

embedding/cpp/flutter_service_app.cc
```diff
@@ -0,0 +1,173 @@
...
+  handle = FlutterRunEngine(engine_prop);
+  if (!handle) {
+    TizenLog::Error("Could not launch a Flutter Service application.");
+    return false;
+  }
+
+  return true;
+}
+
+void FlutterApp::OnResume() {
```
pkosko, line 56: most of this file is a duplication of flutter_app.cc maybe we can consider using inheritance to reduce redundancy of code. Is it possible or there are some contraindications?

pkosko (commented):
Few questions/doubts to check.

templates/app/cpp/tizen-manifest-service.xml.tmpl
rwalczyna: It is added automatically when creating service application by Tizen Studio.

templates/app/cpp/tizen-manifest-service.xml.tmpl
rwalczyna: AFAIK native application does not have `background support` selector. And it looks like `background-category` is enough.

embedding/cpp/flutter_service_app.cc
rwalczyna: I tried to make everything as simple as possible. But I will check if there is any options of reducing code.

rwalczyna:
> The type `FlutterApp` is ambiguous. Wouldn't it be better to create an explicit type `FlutterServiceApp` (either from scratch or by extending `FlutterApp`)? Maybe your intention was to share the same files under `templates/app/cpp` across the different app types

@swift-kim, yes, I wanted to share files that don't need modifying. I will try approach you suggested.

swift-kim:
@rwalczyna Thanks!

Please also refer to https://github.com/flutter-tizen/flutter-tizen/pull/105 and follow the updated APIs.

The `virtual` keyword has been added to `FlutterApp.Run()` so you can override it from `FlutterServiceApp` if you're extending the `FlutterApp` class. You can also change the access level of `ParseEngineArgs()` to `protected` so that you can re-use in your class (I forgot to update it).

embedding/cpp/flutter_app.cc
```diff
@@ -193,3 +193,92 @@ void FlutterApp::ParseEngineArgs() {
...
+                                  FlutterApp *app = (FlutterApp *)data;
+                                  app->OnLanguageChanged(e);
+                                },
+                                this);
+  service_app_add_event_handler(&handler, APP_EVENT_REGION_FORMAT_CHANGED,
+                                [](app_event_info_h e, void *data) {
+                                  FlutterApp *app = (FlutterApp *)data;
+                                  app->OnRegionFormatChanged(e);
+                                },
+                                this);
```
swift-kim, line 236: Could you re-format this part?

```suggestion
  service_app_add_event_handler(
      &handler, APP_EVENT_LOW_MEMORY,
      [](app_event_info_h e, void *data) {
        FlutterApp *app = (FlutterApp *)data;
        app->OnLowMemory(e);
      },
      this);
  service_app_add_event_handler(
      &handler, APP_EVENT_LOW_BATTERY,
      [](app_event_info_h e, void *data) {
        FlutterApp *app = (FlutterApp *)data;
        app->OnLowBattery(e);
      },
      this);
  service_app_add_event_handler(
      &handler, APP_EVENT_LANGUAGE_CHANGED,
      [](app_event_info_h e, void *data) {
        FlutterApp *app = (FlutterApp *)data;
        app->OnLanguageChanged(e);
      },
      this);
  service_app_add_event_handler(
      &handler, APP_EVENT_REGION_FORMAT_CHANGED,
      [](app_event_info_h e, void *data) {
        FlutterApp *app = (FlutterApp *)data;
        app->OnRegionFormatChanged(e);
      },
      this);
```

embedding/cpp/include/flutter_app.h
```diff
@@ -52,8 +54,19 @@ class FlutterApp : public flutter::PluginRegistry {
   // The Flutter engine instance handle.
   FlutterDesktopEngineRef handle;
 
- private:
+ protected:
   void ParseEngineArgs();
 };
 
+// The app base class for headless execution.
+class FlutterServiceApp : public FlutterApp {
```
swift-kim, line 62: I'm not sure, but would it be better to move this class into a separate header?

embedding/cpp/flutter_app.cc
rwalczyna: Thanks, I was using VSCode with Google style to format code, but it turned out that I had clang-format-6.0 installed. After updating it to version 10, formatting also changed.

embedding/cpp/flutter_service_app.cc
```diff
@@ -0,0 +1,105 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "include/flutter_service_app.h"
+
+#include <cassert>
+#include <cerrno>
+#include <fstream>
+
+#include "include/flutter_app.h"
```
swift-kim, line 11: Please remove these unnecessary includes.

embedding/cpp/include/flutter_app.h
```diff
@@ -6,6 +6,7 @@
 #define FLUTTER_TIZEN_EMBEDDING_CPP_INCLUDE_FLUTTER_APP_H_
 
 #include <app.h>
+
```
swift-kim, line 9: Accidentally added line break.

embedding/cpp/include/flutter_service_app.h
```diff
@@ -0,0 +1,25 @@
+// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef FLUTTER_TIZEN_EMBEDDING_CPP_INCLUDE_FLUTTER_SERVICE_APP_H_
+#define FLUTTER_TIZEN_EMBEDDING_CPP_INCLUDE_FLUTTER_SERVICE_APP_H_
+
+#include <flutter/plugin_registry.h>
```
swift-kim, line 8:
```suggestion

```

swift-kim (commented):
Otherwise looks good to me!

embedding/cpp/flutter_app.cc
swift-kim: @pkosko Uhm, adding a code style checker for the engine and plugins repo is absolutely fine, but this repo only contains a small portion of native code and I'm not sure we need such checker for this repo. @flutter-tizen/maintainers How do you think?

swift-kim: I added this page to our wiki: https://github.com/flutter-tizen/flutter-tizen/wiki/Style-guide

Please let me know if I missed something.

swift-kim (approved):

PR #98, opened 2021-04-30
Compile native plugins into a single shared object

### LINKED ISSUE #90: Destroy the plugin properly when exiting the app
According to my analysis, when using multiple plug-ins as shown below, only the last registered plugin is destroyed normally when exiting app. This is causing a crash at the end of the app. 

I guess that there are something wrong with implementations that deals with plugins in Tizen Embedder

```cs
//
// Generated file. Do not edit.
//
using System;
using System.Runtime.InteropServices;
using Tizen.Flutter.Embedding;

namespace Runner
{
    internal class GeneratedPluginRegistrant
    {
        [DllImport("camera_plugin.so")]
        public static extern void CameraPluginRegisterWithRegistrar(IntPtr registrar);
        [DllImport("integration_test_plugin.so")]
        public static extern void IntegrationTestPluginRegisterWithRegistrar(IntPtr registrar);
        [DllImport("video_player_tizen_plugin.so")]
        public static extern void VideoPlayerTizenPluginRegisterWithRegistrar(IntPtr registrar);

        public static void RegisterPlugins(FlutterApplication app)
        {
            CameraPluginRegisterWithRegistrar(app.GetPluginRegistrar("CameraPlugin"));
            IntegrationTestPluginRegisterWithRegistrar(app.GetPluginRegistrar("IntegrationTestPlugin"));
            VideoPlayerTizenPluginRegisterWithRegistrar(app.GetPluginRegistrar("VideoPlayerTizenPlugin"));
        }
    }
}

```

swift-kim:
[Before]
- (Native) Plugins are compiled as part of the runner executable.
- (.NET) Plugins are built separately into each .so and loaded by the runner (managed DLL) via P/Invoke. This results in the error stated by #90, because each plugin creates its own `PluginRegistrarManager` instance and tries to register its destruction handler to the embedder.

[After]
- Plugins are compiled into a single file `libflutter_plugins.so` regardless of the app type.

Changelog:
- Re-write the `TizenPlugins` build target
- Update `DotnetTpk.build()` and `NativeTpk.build()` accordingly and clean up
- Remove the `TizenLibrary` class and merge it with `TizenPlugin`
- Change the TPK output directory (from `build/tizen` to `build/tizen/tpk`)
- Wrap the plguin class in the plugin template with an unnamed namespace to prevent potential naming conflicts
- Replace `__FILE__` with `__MODULE__` in `templates/plugin/cpp/src/log.h.tmpl`

Fixes #90.

### DIFF
--- a/lib/tizen_build_target.dart
+++ b/lib/tizen_build_target.dart
@@ -103,9 +103,8 @@ abstract class TizenAssetBundle extends Target {
 
 /// Source: [DebugAndroidApplication] in `android.dart`
 class DebugTizenApplication extends TizenAssetBundle {
-  DebugTizenApplication(this.project, this.buildInfo);
+  DebugTizenApplication(this.buildInfo);
 
-  final FlutterProject project;
   final TizenBuildInfo buildInfo;
 
   @override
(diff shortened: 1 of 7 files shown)


bbrto21 (approved):
LGTM! Thanks for fixing #90!

PR #101, opened 2021-05-04
Fix TizenPlugins unexpectedly skipping builds

swift-kim:
Currently there is a bug that ignores any code changes. The map function doesn't work as I expected.


bbrto21 (approved):
LGTM, Thanks for the quick fix!

PR #102, opened 2021-05-04
A workaround for library loading issue on TV emulator

### LINKED ISSUE #50: [TV emulator] "Unhandled exception: System.DllNotFoundException" on first launch of an app
```
I/ConsoleMessage(10510): FlutterApplication.cs: OnCreate(54) > switches: --enable-dart-profiling --enable-checked-mode
E/ConsoleMessage(10510): FlutterApplication.cs: Run(39) > Unhandled exception: System.DllNotFoundException: Unable to load shared library 'flutter_engine.so' or one of its dependencies. In order to help diagnose loading problems, consider setting the LD_DEBUG environment variable: libflutter_engine.so: cannot open shared object file: No such file or directory
E/ConsoleMessage(10510):    at Tizen.Flutter.Embedding.Interop.FlutterEngineRunsAOTCompiledDartCode()
E/ConsoleMessage(10510):    at Tizen.Flutter.Embedding.FlutterApplication.OnCreate()
E/ConsoleMessage(10510):    at Runner.App.OnCreate()
E/ConsoleMessage(10510):    at Tizen.Applications.CoreBackend.UICoreBackend.OnCreateNative(IntPtr data)
E/ConsoleMessage(10510):    at Tizen.Applications.CoreBackend.UICoreBackend.Run(String[] args)
E/ConsoleMessage(10510):    at Tizen.Applications.CoreApplication.Run(String[] args)
E/ConsoleMessage(10510):    at Tizen.Applications.CoreUIApplication.Run(String[] args)
E/ConsoleMessage(10510):    at Tizen.Flutter.Embedding.FlutterApplication.Run(String[] args)
E/ConsoleMessage(10510):    at Runner.App.Main(String[] args)
```

The reason is still unknown but it may have something to do with the Linux kernel used by TV emulators.

A temporary solution:
1. Change the `api-version` value in `tizen/tizen-manifest.xml` file to something different (e.g. 4.0 → 6.0)
2. Launch the app (`flutter-tizen run`)
3. Change the value back to the original value (the error will not be generated on successive launches after the app launches successfully at least once)

swift-kim:
This reverts #34 and applies a solution described by https://github.com/flutter-tizen/flutter-tizen/issues/50#issuecomment-831675478.

Fixes #50.

### DIFF
--- a/lib/tizen_device.dart
+++ b/lib/tizen_device.dart
@@ -267,7 +267,14 @@ class TizenDevice extends Device {
       }
     }
     _logger.printTrace('Installing TPK.');
-    if (!await installApp(package)) {
+    if (await installApp(package)) {
+      // On TV emulator, a tpk must be installed twice if it's being installed
+      // for the first time in order to prevent a library loading error.
+      // Issue: https://github.com/flutter-tizen/flutter-tizen/issues/50
+      if (!wasInstalled && usesSecureProtocol && await isLocalEmulator) {
+        await installApp(package);
+      }
+    } else {
       _logger.printTrace('Warning: Failed to install TPK.');
       if (wasInstalled) {
         _logger.printStatus('Uninstalling old version...');
--- a/embedding/csharp/FlutterApplication.cs
+++ b/embedding/csharp/FlutterApplication.cs
@@ -64,13 +64,6 @@ namespace Tizen.Flutter.Embedding
                 switches_count = (uint)switches.Length,
             };
 
-            // This check is not actually required, but we want to make sure that libflutter_engine.so is loaded
-            // before a call to FlutterCreateWindow() is made to avoid library loading issues on TV emulator.
-            if (FlutterEngineRunsAOTCompiledDartCode())
-            {
-                InternalLog.Debug(LogTag, "Running in AOT mode.");
-            }
-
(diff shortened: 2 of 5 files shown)


lib/tizen_device.dart
```diff
@@ -267,7 +267,14 @@ class TizenDevice extends Device {
       }
     }
     _logger.printTrace('Installing TPK.');
-    if (!await installApp(package)) {
+    if (await installApp(package)) {
+      // On TV emulator, a tpk must be installed twice if it's being installed
+      // for the first time in order to prevent a library loading error.
+      // Issue: https://github.com/flutter-tizen/flutter-tizen/issues/50
+      if (!wasInstalled && usesSecureProtocol && await isLocalEmulator) {
+        await installApp(package);
```
bbrto21, line 275: Interesting and it's hard to understand easily. does this solve the loading problem that occur when launching fresh app?

swift-kim: Yeah, it's a hack and I don't understand why it works either. We may have to contact the TV SDK team if you think this is not a right way to resolve the issue.

bbrto21: I think it's gonna take a lot of time. Let's just keep moving on

bbrto21 (approved):
LGTM

PR #103, opened 2021-05-04
Reference Tizen.Flutter.Embedding via ProjectReference

swift-kim:
Reference Tizen.Flutter.Embedding by project reference instead of package reference in `Runner.csproj`.

In this way, developers can easily compile and debug C# code (the runner and Tizen.Flutter.Embedding) using the .NET CLI and IDEs (such as Visual Studio).

Notable changes:
- Implement `ensureReadyForTizenTooling()`
- `TizenProject.ensureReadyForPlatformSpecificTooling()` creates a `Runner.csproj.user` file next to `Runner.csproj` which is managed by Visual Studio and flutter-tizen and not version controlled
- Remove `NuGet.Config`
- Update `_knownPlugins` with newly added plugins

Although this change does not break existing projects, it is recommended to re-generate Tizen projects using `flutter-tizen create`. The use of `PackageReference` and `FlutterEmbeddingVersion` in `Runner.csproj` is now deprecated.

### DIFF
--- a/lib/tizen_build_target.dart
+++ b/lib/tizen_build_target.dart
@@ -431,6 +431,20 @@ class DotnetTpk {
     // 2021. Keep the value up to date until majority of projects are migrated
     // to use ProjectReference.
     const String embeddingVersion = '1.7.0';
+    final bool migrated = !tizenProject.projectFile
+        .readAsStringSync()
+        .contains(r'$(FlutterEmbeddingVersion)');
+    if (!migrated) {
+      final Function relative = environment.fileSystem.path.relative;
+      environment.logger.printStatus(
+        'The use of PackageReference in ${tizenProject.projectFile.basename} is deprecated. '
+        'To migrate your project, run:\n'
+        '  rm ${relative(tizenProject.projectFile.path)}\n'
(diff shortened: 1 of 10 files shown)


lib/tizen_build_target.dart
```diff
@@ -427,7 +427,9 @@ class DotnetTpk {
           (File lib) => lib.copySync(libDir.childFile(lib.basename).path));
     }
 
-    // Keep this value in sync with the latest published nuget version.
+    // TODO(swift-kim): This property is used by projects created before May
+    // 2021. Keep the value up to date until majority of projects are migrated
```
bbrto21, line 431: How about using a annotation(```@deprecated```)?

swift-kim: The annotation is not applicable here so I added some warning message instead. How do you think?
```
The use of PackageReference in Runner.csproj is deprecated. To migrate your project, run:
  rm tizen/Runner.csproj
  flutter-tizen create .
```

bbrto21: Oh, I think that's a good idea! Thank you.

bbrto21 (approved):
LGTM

PR #105, opened 2021-05-10
Update for the embedder API changes

swift-kim:
- Update `Interop.cs` and `flutter_app.cc` correspondingly to the changes in `flutter_tizen.h` (https://github.com/flutter-tizen/engine/pull/83)
- Move PInvoke-related files into `Interop` directory
- Implement some new functions and types (mostly for later use)
  - `FlutterDesktopMessengerSend`
  - `FlutterDesktopMessengerSendWithReply`
  - `FlutterDesktopMessengerSendResponse`
  - `FlutterDesktopMessengerSetCallback`
  - `FlutterDesktopBinaryReply`
  - `FlutterDesktopMessage`
  - `FlutterDesktopMessageCallback`
  - `FlutterDesktopMessageResponse`
  - `FlutterDesktopMessenger`
  - `FlutterDesktopPluginRegistrar`

Should be merged after https://github.com/flutter-tizen/engine/pull/83 is merged.


embedding/csharp/FlutterApplication.cs
```diff
@@ -64,7 +64,7 @@ protected override void OnCreate()
                 switches_count = (uint)switches.Length,
```
swift-kim, line 64: Note: The C# equivalent of `size_t` in C is `nuint` (available in C# 9.0) or `IntPtr` (technically identical), but somehow marshaling from `uint` still works fine on arm64 devices. I don't want to increase the C# and .NET SDK version requirement right now so I'll leave as it is.

PR #106, opened 2021-05-10
Link libstdc++ statically

### LINKED ISSUE #71: A plugin built in debug mode cannot be loaded on Tizen 5.5
A crash occurs when

- An app tries to load and register a native plugin (in `RegisterPlugins()`)
- The plugin has a dependency on glibc++ 3.4.**26**

The dependency on glibc++ 3.4.26 can be made by for example,

1. Using C++17 APIs explicitly (such as `std::filesystem`)
1. Building any code which makes use of `std::string.data()` (only in debug mode)

The second case is easily reproducible by running the [current](https://github.com/flutter-tizen/plugins/tree/b7b2f230195b7f11676a03f45f8ed438f94cfc16/packages/permission_handler) permission_handler plugin example on any Tizen 5.5 device or emulator. There is an occurrence of `std::string.data()` in the plugin code:

https://github.com/flutter-tizen/plugins/blob/b7b2f230195b7f11676a03f45f8ed438f94cfc16/packages/permission_handler/tizen/src/permission_handler_tizen_plugin.cc#L42

Symbol table:
```sh
$ objdump -T libpermission_handler_tizen_plugin.so | grep GLIBCXX_3.4
00000000      DF *UND*  00000000  GLIBCXX_3.4.21 _ZNKSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEE7compareERKS4_
00000000      DF *UND*  00000000  GLIBCXX_3.4 _ZSt29_Rb_tree_insert_and_rebalancebPSt18_Rb_tree_node_baseS0_RS_
00000000      DF *UND*  00000000  GLIBCXX_3.4 _ZNSaIcED2Ev
00000000      DF *UND*  00000000  GLIBCXX_3.4 _ZSt9terminatev
00000000      DF *UND*  00000000  GLIBCXX_3.4.21 _ZNKSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEE7compareEPKc
00000000      DF *UND*  00000000  GLIBCXX_3.4 _ZSt17__throw_bad_allocv
00000000      DF *UND*  00000000  GLIBCXX_3.4.14 _ZSt25__throw_bad_function_callv
00000000      DF *UND*  00000000  GLIBCXX_3.4.21 _ZNKSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEE5c_strEv
00000000      DF *UND*  00000000  GLIBCXX_3.4.21 _ZNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEC1ERKS4_
00000000      DF *UND*  00000000  GLIBCXX_3.4 _ZSt4endlIcSt11char_traitsIcEERSt13basic_ostreamIT_T0_ES6_
00000000      DF *UND*  00000000  GLIBCXX_3.4 _ZSt20__throw_length_errorPKc
00000000      DF *UND*  00000000  GLIBCXX_3.4.26 _ZNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEE4dataEv  <= this
00000000      DF *UND*  00000000  GLIBCXX_3.4.21 _ZNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEC1EOS4_
...
```
Error:
```
E/ConsoleMessage(27795): FlutterApplication.cs: Run(39) > Unhandled exception: System.DllNotFoundException: Unable to load shared library 'permission_handler_tizen_plugin.so' or one of its dependencies. In order to help diagnose loading problems, consider setting the LD_DEBUG environment variable: libpermission_handler_tizen_plugin.so: cannot open shared object file: No such file or directory
E/ConsoleMessage(27795):    at Runner.GeneratedPluginRegistrant.PermissionHandlerTizenPluginRegisterWithRegistrar(IntPtr registrar)
E/ConsoleMessage(27795):    at Runner.GeneratedPluginRegistrant.RegisterPlugins(FlutterApplication app)
```

As a workaround we can simply replace it with `std::string.c_str()`. However, more deeper investigation is needed to prevent any future errors.

swift-kim:
Link libstdc++ statically when building apps or plugins for Tizen 4.0 and 5.5, to remove dependency on GLIBCXX_3.4.26 and use C++17 features without limitations.

Fixes #71.

Notes:
- Is it safe to statically  link libstdc++ (especially for our case)? - Yes, I couldn't find any reason not to.
- This change slightly increases the tpk and storage size of an app (by 0.4 MB and 1 MB respectively).
- The change is not applied when targeting Tizen 6.0+, because libstdc++.so on Tizen 6.0+ is already compatible with C++17.
- The change is not applied when targeting arm64, because there's only Tizen 6.0+ runtime for arm64.

### DIFF
--- a/embedding/cpp/flutter_app.cc
+++ b/embedding/cpp/flutter_app.cc
@@ -7,7 +7,7 @@
 #include <system_info.h>
 
 #include <cassert>
-#include <filesystem>
+#include <cerrno>
 #include <fstream>
 
 #include "tizen_log.h"
@@ -177,21 +177,18 @@ void FlutterApp::ParseEngineArgs() {
                         std::string(app_id) + ".rpm");
   free(app_id);
 
-  if (!std::filesystem::exists(temp_path)) {
+  std::ifstream file(temp_path);
+  if (!file.is_open()) {
     return;
   }
-  std::ifstream file(temp_path);
-  file.exceptions(std::ifstream::failbit | std::ifstream::badbit);
-
-  try {
-    std::string line;
-    while (!file.eof() && std::getline(file, line)) {
-      TizenLog::Info("Enabled: %s", line.c_str());
-      engine_args.push_back(line.c_str());
-    }
-    std::filesystem::remove(temp_path);
-  } catch (const std::exception &ex) {
-    TizenLog::Warn("Error while processing a file: %s", ex.what());
+  std::string line;
+  while (file && !file.eof() && std::getline(file, line)) {
+    TizenLog::Info("Enabled: %s", line.c_str());
+    engine_args.push_back(line.c_str());
   }
   file.close();
+
+  if (remove(temp_path.c_str()) != 0) {
+    TizenLog::Warn("Error removing file: %s", strerror(errno));
+  }
 }
--- a/embedding/cpp/flutter_app.cc
+++ b/embedding/cpp/flutter_app.cc
@@ -7,7 +7,7 @@
 #include <system_info.h>
 
 #include <cassert>
-#include <cerrno>
+#include <filesystem>
 #include <fstream>
 
 #include "tizen_log.h"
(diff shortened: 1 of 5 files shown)


bbrto21 (approved):
I researched and tested a few things related to this. I can't find any problem. so I approve this PR.
but, it needs to be reviewed by others memeber. I am not an expert in this area.

HakkyuKim (approved):
I also couldn't find any issue with this patch. It solves the problem elegantly. LGTM.

PR #112, opened 2021-05-14
Add a guide for setting up RPi devices

swift-kim:


doc/configure-device.md
```diff
@@ -36,6 +36,98 @@ To debug Flutter apps on your Tizen device (watch or TV), you need to configure
...
+   The filesystem on /dev/sdc5 is now 6456832 (4k) blocks long.
+   ```
+
+1. Done. Detach the SD card and insert to the Pi device.
+
+### Connect
+
+1. Connect the Pi device to the same network as your host PC using an Ethernet or Wi-Fi (built-in or extra USB dongle) connection.
+
+   Alternatively, you can use some extra hardware (USB/PWR splitter) to connect Pi directly to the host PC in OTG (USB slave) mode.
```
pwasowski2, line 121: I'd add a note, that when we connect a device over USB, we don't have to issue `sdb connect ...` - it may not be so obvious to devs new to Tizen ;)

```suggestion
  > Alternatively, you can use some extra hardware (USB/PWR splitter) to connect Pi directly to the host PC in OTG (USB slave) mode. The Tizen device will be ready to use just after plugging the USB - the next step should be skipped.
```

pwasowski2 (commented):
Very good tutorial for newcomers 👏 Consider incorporating one suggestion ;)

doc/configure-device.md
swift-kim: Thank you!

pwasowski2 (approved):
LGTM :)

PR #113, opened 2021-05-14
Fix create command for C# application

rwalczyna:
Creating C# application caused failure because of not existing app
directory.


pkosko (approved):
Commit helps to fix described issue there is no failure when creating C# templated application.

swift-kim (approved):

PR #117, opened 2021-05-19
Clarify install steps for macOS

swift-kim:
macOS is using zsh as the default shell since macOS Catalina (the previous version of macOS Big Sur). I think now it's time to update our documentation.


swift-kim:
Force-pushed after merge to fix a typo in line 31.

PR #120, opened 2021-05-23
Properly escape spaces in user paths

swift-kim:
- Update the installation guide.
  - Wrap with `""` when setting PATH.
  - Also move the Windows WSL guide to the wiki.
- Update extra options for the Tizen native builder.
  - Wrap paths with `""`.
  - Note that this does not fully fix the problem stated by https://github.com/flutter-tizen/plugins/issues/104, because the paths in `USER_SRCS` are still not escaped. Wrapping with `""` wouldn't help here because Make (which is used by the Tizen native builder) doesn't support it. I couldn't find any non-tricky way to work around this issue. My conclusion is that Make simply cannot handle spaces.
    - https://stackoverflow.com/questions/17965806/how-do-i-handle-files-with-spaces-in-my-makefile
    - https://stackoverflow.com/questions/9838384/can-gnu-make-handle-filenames-with-spaces


bbrto21 (approved):
LGTM

PR #123, opened 2021-05-27
Simplify the __MODULE__ macro definition

swift-kim:
As suggested in https://github.com/flutter-tizen/plugins/pull/111#pullrequestreview-667883789.


rwalczyna (approved):

bbrto21 (approved):
LGTM

Issue #134, opened 2021-06-08
Flutter Service Application crashes when using statically linked standard library


rwalczyna:
Flutter Service Application crashes when `-static-libstdc++` flag is used (#106). It looks like [getline](https://github.com/flutter-tizen/flutter-tizen/blob/20007aa24f83ff21ff3f5314410f16e448b0d75c/embedding/cpp/flutter_app.cc#L185) in [ParseEngineArgs](https://github.com/flutter-tizen/flutter-tizen/blob/20007aa24f83ff21ff3f5314410f16e448b0d75c/embedding/cpp/flutter_app.cc#L170) causes bad_cast exception

```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0xb6e2b788 in __GI_abort () at abort.c:79
#2  0x2a0c0f74 in __gnu_cxx::__verbose_terminate_handler() ()
#3  0x2a03d608 in __cxxabiv1::__terminate(void (*)()) ()
#4  0x2a03d67c in std::terminate() ()
#5  0x2a03d82c in __cxa_throw ()
#6  0x2a008a14 in std::__throw_bad_cast() ()
#7  0x2a086980 in std::istream::getline(char*, int) ()
#8  0x2a00a5ac in FlutterApp::ParseEngineArgs (this=0x2a10a290) at /home/r.walczyna/git/flutter/flutter-tizen/embedding/cpp/flutter_app.cc:189
#9  0x2a00d87c in FlutterServiceApp::OnCreate (this=0x2a10a290) at /home/r.walczyna/git/flutter/flutter-tizen/embedding/cpp/flutter_service_app.cc:79
#10 0x2a009a60 in App::OnCreate (this=0x2a10a290) at src/runner.cc:8
#11 0x2a00db18 in FlutterServiceApp::Run(int, char**)::$_0::operator()(void*) const (this=0x2a10a290, data=0x2a10a290) at /home/r.walczyna/git/flutter/flutter-tizen/embedding/cpp/flutter_service_app.cc:13
#12 0x2a00dae0 in FlutterServiceApp::Run(int, char**)::$_0::__invoke(void*) (data=0x2a10a290) at /home/r.walczyna/git/flutter/flutter-tizen/embedding/cpp/flutter_service_app.cc:11
#13 0xb6f7843c in ?? () from /lib/libappcore-agent.so.1
[...]

PR #135, opened 2021-06-09
Fix service application crash on start

rwalczyna:
ParseEngineArgs was causing crash because of statically linked c++ standard libraries.

Details: #134

### DIFF
--- a/embedding/cpp/flutter_app.cc
+++ b/embedding/cpp/flutter_app.cc
@@ -179,7 +179,6 @@ void FlutterApp::ParseEngineArgs() {
 
   auto file = fopen(temp_path.c_str(), "r");
   if (!file) {
-    TizenLog::Error("file is closed");
     return;
   }
   char *line = nullptr;
--- a/embedding/cpp/flutter_app.cc
+++ b/embedding/cpp/flutter_app.cc
@@ -177,16 +177,20 @@ void FlutterApp::ParseEngineArgs() {
                         std::string(app_id) + ".rpm");
   free(app_id);
 
-  std::ifstream file(temp_path);
-  if (!file.is_open()) {
+  auto file = fopen(temp_path.c_str(), "r");
+  if (!file) {
+    TizenLog::Error("file is closed");
     return;
   }
-  std::string line;
-  while (file && !file.eof() && std::getline(file, line)) {
-    TizenLog::Info("Enabled: %s", line.c_str());
-    engine_args.push_back(line.c_str());
+  char *line = nullptr;
+  size_t len = 0;
+
+  while (getline(&line, &len, file) > 0) {
+    TizenLog::Info("Enabled: %s", line);
+    engine_args.push_back(line);
   }
-  file.close();
+  free(line);
+  fclose(file);
 
   if (remove(temp_path.c_str()) != 0) {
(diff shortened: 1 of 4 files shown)


pkosko (approved):
LGTM

embedding/cpp/flutter_service_app.cc
```diff
@@ -98,3 +98,34 @@ bool FlutterServiceApp::OnCreate() {
...
   return true;
 }
+
+void FlutterServiceApp::ParseEngineArgs() {
+  char *app_id;
+  if (app_get_id(&app_id) != 0) {
+    TizenLog::Warn("App id is not found.");
+    return;
+  }
+  std::string temp_path("/home/owner/share/tmp/sdk_tools/" +
```
bwikbs, line 108: Maybe that path will change later, so how about subtracting it with a variable or define? And I think a method other than hard coding would be better...

swift-kim: @bwikbs https://github.com/flutter-tizen/flutter-tizen/pull/19 explains why we use this dirty hack.

The path should never change because it's the only directory sdb has a read/write access to on TV. If you're to change the hard-coded value to a constant, pls make sure to make a consistent change in the C# embedding as well.

swift-kim (commented):
Why don't you just update `FlutterApp::ParseEngineArgs()` to use fopen? I don't think we need two functions that effectively do the same thing.

embedding/cpp/flutter_service_app.cc
bwikbs: @swift-kim Thanks for letting me know! 
What do you think of the aul parameter way? Shall we give it a try?

swift-kim: Uhm, #19 explains why we can't use other way than the current way. We used to use \_\_DLP_UNIT_TEST_ARG__ to pass arguments to launched apps before the PR but that method didn't work for TV devices.

bwikbs: Ah, I didn't think of commercial TV. Thanks for letting me know.

embedding/cpp/flutter_app.cc
```diff
@@ -177,16 +177,20 @@ void FlutterApp::ParseEngineArgs() {
                         std::string(app_id) + ".rpm");
   free(app_id);
 
-  std::ifstream file(temp_path);
-  if (!file.is_open()) {
+  auto file = fopen(temp_path.c_str(), "r");
+  if (!file) {
+    TizenLog::Error("file is closed");
```
pkosko, line 182: Maybe add a path to this log to point exact file which opening failed? Also probably "Failed to open file xxxx" will be more accurate.

rwalczyna: I would consider removing this log, because this file exists only on first run.

swift-kim: @rwalczyna is right. Please keep everything in sync with the C# implementation:

https://github.com/flutter-tizen/flutter-tizen/blob/20007aa24f83ff21ff3f5314410f16e448b0d75c/embedding/csharp/FlutterApplication.cs#L74-L96

swift-kim (approved):

bwikbs (approved):

Issue #137, opened 2021-06-10
Document how to publish apps on the TV app store


swift-kim:
We need to fill the part in https://github.com/flutter-tizen/flutter-tizen/blob/master/doc/publish-app.md.

iandis:
publishing TV app to samsung seller 2.0 seems a little bit confusing for me. I was trying to upload my .tpk and the pre-test failed. It said my tpk needed to have <tv-info> in the tizen-manifest. I tried finding it online but found nothing helpful.

swift-kim:
@iandis It looks like a temporary issue. I'll contact the person in charge in the TV division and let you know when I get a reply.

swift-kim:
@iandis I got a response and they're going to fix the issue in a few days, but you can also try adding the following lines to your tizen-manifest.xml to work around the issue.

```xml
<tv-info api-version="8.8.0">
    <infolink>T-INFOLINK2021-1000</infolink>
</tv-info>
```

iandis:
@swift-kim Thank you very much for the quick response😃!

PR #139, opened 2021-06-14
Add <tv-info> element to tizen-manifest.xml

swift-kim:
- Add the \<tv-info> element to the tizen-manifest.xml template (.NET-only). The TV seller portal automatically rejects app binaries without the element (https://github.com/flutter-tizen/flutter-tizen/issues/137#issuecomment-860364896). The Galaxy Store binary checker will ignore the element.
- Remove trailing newlines from the templates. The tool automatically removes them when building a tpk.

Please refer to the [internal guide](http://wiki.vd.sec.samsung.net/pages/viewpage.action?spaceKey=TIZ&title=dotnet-inhouse+type) for detailed information about the element.


rwalczyna:
New lines are deleted by [updateManifest](https://github.com/flutter-tizen/flutter-tizen/blob/c01b4b5264cf317efc412c6a69d85ee1f49deadf/lib/tizen_builder.dart#L237) function. I think that it would be better to keep the new lines in template files and also add them in `updateManifest`.

swift-kim:
@rwalczyna Alright. I'll change `updateManifest()` as you suggested.

rwalczyna (approved):

Issue #142, opened 2021-06-16
Tizen toolchain - develop for Tizen devices (the doctor check crashed)


EbramTawfik:
I am installing flutter-tizen and when I run `flutter-tizen doctor -v` I am getting this error
![image](https://user-images.githubusercontent.com/12788360/122273847-c1fe0b00-ce96-11eb-8795-9a78dee59a31.png)

`Tizen toolchain - develop for Tizen devices (the doctor check crashed)
    X Due to an error, the doctor check did not complete. If the error message below is not helpful, please let us know
      about this issue at https://github.com/flutter/flutter/issues.
    X NoSuchMethodError: The getter 'length' was called on null.
      Receiver: null
      Tried calling: length
    • #0      Object.noSuchMethod (dart:core-patch/object_patch.dart:54:5)
      #1      double._parse (dart:core-patch/double_patch.dart:95:19)
      #2      double.tryParse (dart:core-patch/double_patch.dart:118:45)
      #3      TizenValidator.validate (package:flutter_tizen/tizen_doctor.dart:88:38)
      #4      asyncGuard.<anonymous closure> (package:flutter_tools/src/base/async_guard.dart:111:32)
      #5      asyncGuard.<anonymous closure> (package:flutter_tools/src/base/async_guard.dart:109:18)
      #6      _rootRun (dart:async/zone.dart:1354:13)
      #7      _CustomZone.run (dart:async/zone.dart:1258:19)
      #8      _runZoned (dart:async/zone.dart:1788:10)
      #9      runZonedGuarded (dart:async/zone.dart:1776:12)
      #10     runZoned (dart:async/zone.dart:1707:12)
      #11     asyncGuard (package:flutter_tools/src/base/async_guard.dart:109:3)
      #12     Doctor.startValidatorTasks (package:flutter_tools/src/doctor.dart:203:9)
      #13     Doctor.diagnose (package:flutter_tools/src/doctor.dart:307:47)
      #14     DoctorCommand.runCommand (package:flutter_tools/src/commands/doctor.dart:48:47)
      #15     FlutterCommand.verifyThenRunCommand (package:flutter_tools/src/runner/flutter_command.dart:1157:18)
      <asynchronous suspension>
      #16     FlutterCommand.run.<anonymous closure> (package:flutter_tools/src/runner/flutter_command.dart:1009:27)
      <asynchronous suspension>
[...]

PR #144, opened 2021-06-22
Fix doctor crash when validating Tizen Studio version

swift-kim:
Fixes the doctor command crash when the `sdk.version` file is corrupted.

This fixes the error in #142, but doesn't really help fixing the root cause of the error.

### DIFF
--- a/lib/tizen_doctor.dart
+++ b/lib/tizen_doctor.dart
@@ -89,16 +89,14 @@ class TizenValidator extends DoctorValidator {
       return ValidationResult(ValidationType.missing, messages);
     }
 
-    if (tizenSdk.sdkVersion == null) {
+    final double sdkVersion = double.tryParse(tizenSdk.sdkVersion ?? '');
+    if (sdkVersion == null) {
       messages.add(const ValidationMessage.error(
         'Unknown Tizen Studio version.\n'
-        'The version file is corrupted. Consider updating or reinstalling Tizen Studio.',
+        'The version file is missing or corrupted. Consider updating or reinstalling Tizen Studio.',
       ));
       return ValidationResult(ValidationType.missing, messages);
-    }
-
-    final double sdkVersion = double.tryParse(tizenSdk.sdkVersion);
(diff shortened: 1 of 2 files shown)


lib/tizen_doctor.dart
```diff
@@ -89,18 +89,27 @@ class TizenValidator extends DoctorValidator {
       return ValidationResult(ValidationType.missing, messages);
     }
 
-    final double sdkVersion = double.tryParse(tizenSdk.sdkVersion) ?? 0;
-    if (sdkVersion < 4.0) {
+    if (tizenSdk.sdkVersion == null) {
```
WonyoungChoi, line 92: If `tizenSdk.sdkVersion` cannot be parsed, it is better to handle it here. 
```suggestion
    final double sdkVersion = double.tryParse(tizenSdk.sdkVersion ?? '');
    if (sdkVersion == null) {
```

lib/tizen_doctor.dart
```diff
@@ -89,18 +89,27 @@ class TizenValidator extends DoctorValidator {
...
+      messages.add(const ValidationMessage.error(
+        'Unknown Tizen Studio version.\n'
+        'The version file is corrupted. Consider updating or reinstalling Tizen Studio.',
+      ));
+      return ValidationResult(ValidationType.missing, messages);
+    }
+
+    final double sdkVersion = double.tryParse(tizenSdk.sdkVersion);
+    if (sdkVersion == null || sdkVersion < 4.0) {
```
WonyoungChoi, line 101:
```suggestion
    if (sdkVersion < 4.0) {
```

lib/tizen_doctor.dart
swift-kim: I updated the code.

I personally don't like the hack that makes use of `''`, but this way looks functionally correct.

WonyoungChoi (approved):
Looks good to me

HakkyuKim (approved):
LGTM

PR #147, opened 2021-06-24
Show warning when building a native app with default certificate

swift-kim:
- Show a default certificate warning during native app build.
- Capitalize "TPK"s.

### DIFF
--- a/lib/tizen_build_target.dart
+++ b/lib/tizen_build_target.dart
@@ -521,7 +521,7 @@ class DotnetTpk {
       }
     } else {
       environment.logger.printStatus(
-        'The tpk has been signed with a default certificate. You can create one using Certificate Manager.\n'
+        'The TPK was signed with a default certificate. You can create one using Certificate Manager.\n'
         'https://github.com/flutter-tizen/flutter-tizen/blob/master/doc/install-tizen-sdk.md#create-a-tizen-certificate',
         color: TerminalColor.yellow,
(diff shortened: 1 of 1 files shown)


HakkyuKim:
While changing the code here, how about matching the code with `DonetTpk`? So that it doesn't package again for default certificate.

suggested change:

```dart
// NativeTpk `build()`
if (securityProfile != null) {
    environment.logger
        .printStatus('The $securityProfile profile is used for signing.');
    result = await _processUtils.run(<String>[
        tizenSdk.tizenCli.path,
        'package',
        '-t',
        'tpk',
        if (securityProfile != null) ...<String>['-s', securityProfile],
        '--',
        buildDir.path,
    ]);
    if (result.exitCode != 0) {
        throwToolExit('Failed to generate TPK:\n$result');
    }
} else {
    environment.logger.printStatus(
    'The TPK was signed with a default certificate. You can create one using Certificate Manager.\n'
    'https://github.com/flutter-tizen/flutter-tizen/blob/master/doc/install-tizen-sdk.md#create-a-tizen-certificate',
    color: TerminalColor.yellow,
    );
}
```

HakkyuKim (approved):
@swift-kim explained that `DotnetTpk` and `NativeTpk` use different build commands, the packaging step for `NativeTpk` is actually required after build. My previous comment is no longer relevant. LGTM.

PR #159, opened 2021-07-06
Add support for FlutterDesktopWindowProperties

swift-kim:
Update and refactor embedding classes according to the changes in https://github.com/flutter-tizen/engine/pull/133.

Allows the user app to change the window offset and size before calling `FlutterApplication.OnCreate()`.

Example (App.cs):

```csharp
protected override void OnCreate()
{
    WindowOffsetX = 0;
    WindowOffsetY = 300;
    WindowWidth = 320;
    WindowHeight = 480;

    base.OnCreate();

    GeneratedPluginRegistrant.RegisterPlugins(this);
}
```

Depends on https://github.com/flutter-tizen/engine/pull/133.


bwikbs (approved):
:clap: :clap: :clap: It looks really GOOD!

embedding/cpp/include/flutter_app.h
```diff
@@ -40,20 +40,41 @@ class FlutterApp : public flutter::PluginRegistry {
...
+  int32_t window_width_ = 0;
+
+  // The height of the window, or the maximum height if the value is zero.
+  int32_t window_height_ = 0;
+
+  // Whether the window should have a transparent background or not.
+  bool is_window_transparent_ = false;
+
+  // Whether the window should be focusable or not.
+  bool is_window_focusable_ = true;
```
WonyoungChoi, line 70: Is there any reason not to use the `FlutterDesktopWindowProperties` struct directly?

swift-kim: My intention here was not to expose the embedder API directly to the user as much as possible, because that's why this C++ class is there. If you think using the struct makes the code cleaner, I can make the change, but we will have to think how to make a consistent change in the C# embedding. I don't really want the user to reference the `Interop` class (C#) directly although it's public.

WonyoungChoi: If the window related properties are not going to change much in the future, I think current code is **fine**. And I don't want the `Interop` classes exposed to the public either. If we use the struct directly here, we may need to create new `Properties` class for the public and provide a implicit operator for converting to the Interop struct. (It's a little strange, too.... )

swift-kim: These config values are kind of experimental and subject to change in the future (the user should use them at their own risk). Creating a dedicated class seems to be a right choice, but maybe it's too much for now.

WonyoungChoi: @swift-kim I agree `it's too much for now` :)

WonyoungChoi (approved):

PR #162, opened 2021-07-11
Do not validate the Tizen SDK version when sdk.version is missing

swift-kim:
Issue: The `sdk.version` file is not generated under some circumstances, likely after installing Tizen Studio through the CLI installer.

Changes:
- Let the doctor not complain when the SDK version is not found.
- Use an instance of `Version` to validate the .NET SDK version.


HakkyuKim (approved):
LGTM

WonyoungChoi (approved):
Version class made the code more beautiful! 👍

PR #163, opened 2021-07-14
Install TPK twice when uninstalling and re-installing (TV emulator)

swift-kim:
Currently an app is installed only once on TV emulator when uninstalling and installing (e.g. due to certificate change) an already installed app. This may lead to the notorious library loading error (https://github.com/flutter-tizen/flutter-tizen/issues/50).

- Make sure the app is installed twice.
- Apply https://github.com/flutter/flutter/pull/67936 (change the functions order and un-nest if blocks).
- Check the uninstallation result.
- Use `Version` when comparing the package API version and the device API version.

### DIFF
--- a/lib/tizen_device.dart
+++ b/lib/tizen_device.dart
@@ -251,10 +251,10 @@ class TizenDevice extends Device {
   /// Source: [AndroidDevice.uninstallApp] in `android_device.dart`
   @override
   Future<bool> uninstallApp(TizenTpk app, {String userIdentifier}) async {
-    try {
-      await runSdbAsync(<String>['uninstall', app.id]);
-    } on Exception catch (error) {
-      _logger.printError('sdb uninstall failed: $error');
+    final RunResult result =
+        await runSdbAsync(<String>['uninstall', app.id], checked: false);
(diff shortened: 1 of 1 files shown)


HakkyuKim (approved):
LGTM

lib/tizen_device.dart
```diff
@@ -210,16 +211,53 @@ class TizenDevice extends Device {
...
+    _logger.printTrace('Warning: Failed to install TPK.');
+    if (!wasInstalled) {
+      return false;
+    }
+    _logger.printStatus('Uninstalling old version...');
+    if (!await uninstallApp(app)) {
+      _logger.printError('Error: Uninstalling old version failed.');
+      return false;
+    }
+    if (!await _installApp(app, installTwice: isTvEmulator)) {
```
WonyoungChoi, line 235: Add a trace log for this second try

swift-kim: When the first try fails the user will see

```
Installing TPK.
Installing xxxxxxx.tpk...
Installing TPK failed: (error log)
Warning: Failed to install TPK.
Uninstalling old version...
Installing xxxxxxx.tpk...
```

Isn't this information enough?

PR #167, opened 2021-07-20
Support creating C# service applications

swift-kim:
- Create the `FlutterServiceApplication` class in `Tizen.Flutter.Embedding` and refactor for code sharing.
- Change the directory structure of `templates`.
  - e.g. `templates/app/cpp/service-app` → `templates/service-app/cpp`
- Add the C# service app template.
- Simplify the try-finally block in `create.dart`.
- Remove the `"http://tizen.org/metadata/direct-launch"` metadata from the C# app template. It was initially added to prevent problems like https://github.com/flutter-tizen/engine/issues/58 but it doesn't seem necessary any more.


templates/ui-app/csharp/Runner.csproj
```diff
@@ -0,0 +1,26 @@
+﻿<Project Sdk="Tizen.NET.Sdk/1.1.5">
```
WonyoungChoi, line 1: the latest version of Tizen.NET.Sdk is 1.1.6. 
```suggestion
﻿<Project Sdk="Tizen.NET.Sdk/1.1.6">
```

templates/service-app/csharp/tizen-manifest.xml.tmpl
```diff
@@ -0,0 +1,14 @@
...
+    <profile name="common"/>
+    <service-application appid="{{androidIdentifier}}" exec="Runner.dll" type="dotnet" multiple="false" nodisplay="true" taskmanage="false" api-version="4" auto-restart="false">
+        <label>{{projectName}}</label>
+        <icon>ic_launcher.png</icon>
+        <background-category value="media"/>
+        <metadata key="http://tizen.org/metadata/prefer_dotnet_aot" value="true"/>
+    </service-application>
+    <tv-info api-version="8.8.0">
+        <infolink>T-INFOLINK2021-1000</infolink>
+    </tv-info>
```
WonyoungChoi, line 12: why do we need this information? I think it would be nice to provide a `--tizen-profile` flag for `create` command later.

swift-kim: > why do we need this information?

The information is required by the TV seller portal: https://github.com/flutter-tizen/flutter-tizen/pull/139.

> I think it would be nice to provide a `--tizen-profile` flag for `create` command later. 

There could be many side effects made by this. I'll explain at your seat.

templates/ui-app/csharp/Runner.csproj
swift-kim: A warning will be generated if the app project's SDK version mismatches that of `Tizen.Flutter.Embedding.csproj`.

```
/home/swift/Git/flutter-tizen/embedding/csharp/Tizen.Flutter.Embedding.csproj : warning MSB4240: Multiple versions of the same SDK
"Tizen.NET.Sdk" cannot be specified. The previously resolved SDK version "1.1.5" from location
```

It seems it's still safe and I'll make the change in all csproj files.

WonyoungChoi (approved):
LGTM 👍

swift-kim:
Resolved conflicts with https://github.com/flutter-tizen/flutter-tizen/pull/171.

bbrto21 (approved):

PR #171, opened 2021-07-22
Add support for custom Dart entrypoints

swift-kim:
Usage:

In `main.dart`,
```dart
void main() {
  // I'm not a real main!
}

@pragma('vm:entry-point')
void anotherMain() {
  runApp(MyApp());
}
```

In `runner.cc`,
```cpp
bool OnCreate() {
  dart_entrypoint_ = "anotherMain";
  ...
```

Caveats:

- The app's main Dart file must declare `main()` even if it's unused. It's a requirement by the Dart compiler but not the flutter-tizen tool.
- The use of custom entrypoints may break app builds if the app depends on one or more FFI-based plugins. This is supposed to be a very rare scenario.

Depends on https://github.com/flutter-tizen/engine/pull/152.

Closes https://github.com/flutter-tizen/engine/issues/111.


swift-kim:
Public methods such as `SetEntrypoint()` may be added later. The current support is experimental.

bbrto21 (approved):
I tested this using .Net host and it works.

Issue #168, opened 2021-07-20
Drive command can't detect or display test failures for some plugins.


HakkyuKim:
`flutter-tizen drive` command--which runs integration test--can't detect or display test failures for some plugins, their tests are falsely reported as success.

Patches that would fix the issue:
- [x] Allow dlogs to display on the tool's terminal. (partially done, sometimes drive command exits before returning all dlogs: `video_player` on wearable emulator)
- ~~Receive correct test report from Dart VM service. (detecting test failure)~~
- [x] Inject dart plugins to test runners in tools.
- [x] Update script in https://github.com/flutter-tizen/plugins/pull/150 to use the test command.

HakkyuKim:
I've concluded that the drive command doesn't work very well in the current state. These are the problems:

- drive command reports false result: path_provider(MissingPluginException and FileSystemException due to permission denied are not reported and therefore the command outputs that all tests passed)
- drive command doesn't report result at all: camera
- drive command exits before dlog returns: video_player on emulator

I'm changing my plan to use the test command instead.

As of Flutter 2.2.0, the test command supports running integration tests by looking at the directory name in which the files are located(`flutter test integration_test`).It's not documented yet but it is part of the stable API and from quick testing, this seems like a better option than solving all the problems above. Please see this [design doc](https://docs.google.com/document/d/1jMMZpRAiQC2XTUFnzFsBrWwMsf7KyMuCScYgrJQJOV0/edit?resourcekey=0-ne9HbH1hXCHcIglomNfwtw#heading=h.pub7jnop54q0) and [PR](https://github.com/flutter/flutter/pull/76200) for details.

PR #177, opened 2021-07-27
Support dart plugins in test command

HakkyuKim:
Allows injecting dart plugins for `flutter-tizen test integration_test`. It's sort a hack, I'm open to better suggestions.

Contributes to #168.

You can test with plugins that require dart code: `package_info_plus`, `connectivity_plus`, ...

### DIFF
--- a/lib/commands/test.dart
+++ b/lib/commands/test.dart
@@ -59,7 +59,7 @@ class TizenTestWrapper implements TestWrapper {
       await test.main(args);
       return;
     }
-    
+
     final int filesOptionIndex = args.lastIndexOf('--');
     final List<String> newArgs = args.sublist(0, filesOptionIndex + 1);
     final List<File> testFiles = args
--- a/lib/commands/test.dart
+++ b/lib/commands/test.dart
@@ -9,13 +9,13 @@ import 'dart:async';
 import 'package:file/file.dart';
 import 'package:flutter_tools/src/cache.dart';
 import 'package:flutter_tools/src/commands/test.dart';
-import 'package:flutter_tools/src/test/runner.dart';
-import 'package:flutter_tools/src/test/test_wrapper.dart';
 import 'package:flutter_tools/src/dart/language_version.dart';
 import 'package:flutter_tools/src/dart/package_map.dart';
 import 'package:flutter_tools/src/globals.dart' as globals;
 import 'package:flutter_tools/src/project.dart';
 import 'package:flutter_tools/src/runner/flutter_command.dart';
+import 'package:flutter_tools/src/test/runner.dart';
+import 'package:flutter_tools/src/test/test_wrapper.dart';
 import 'package:package_config/package_config.dart';
 import 'package:test_core/src/executable.dart' as test;
 import 'package:test_core/src/platform.dart' as hack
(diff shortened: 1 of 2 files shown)


HakkyuKim:
On second thought, I don't think it's a good idea to change the code, users will be able to see their code disappearing in IDE. I can add some comments in the "replacing" code, but still it's not a good solution. I'll try again.

swift-kim:
You can either

- pass fake `testFiles` to `test_core` by providing a custom `testWrapper`,
- create a fake entrypoint in `TizenDevice.startApp()` by checking if `mainPath` contains `flutter_test_listener.`, or
- [refactor the current plugin architecture of flutter-tizen and create a build Target which does a similar thing to the upstream's `DartPluginRegistrantTarget`](https://github.com/flutter-tizen/plugins/issues/123#issuecomment-875580942). This is the most preferable way but is a little bit complicated. I'll do some PoC implementation if you agree.

HakkyuKim:
@swift-kim 
Wow, Thank you for the solutions. I haven't checked the last one yet, if it's the most preferable way, then yes I would appreciate the help! If it's something that you're not actively focusing on at the moment, I think the first option looks good as a temporary solution. Please see if the updated commit is what you meant.

lib/commands/test.dart
```diff
@@ -4,13 +4,118 @@
...
 // @dart = 2.8
 
+import 'dart:async';
+
+import 'package:file/file.dart';
+import 'package:flutter_tools/src/cache.dart';
 import 'package:flutter_tools/src/commands/test.dart';
+import 'package:flutter_tools/src/test/runner.dart';
+import 'package:flutter_tools/src/test/test_wrapper.dart';
+import 'package:flutter_tools/src/dart/language_version.dart';
```
swift-kim, line 14: Sort the imports alphabetically.

lib/commands/test.dart
```diff
@@ -4,13 +4,118 @@
...
+      final FlutterProject flutterProject = FlutterProject.current();
+      final PackageConfig packageConfig = await loadPackageConfigWithLogging(
+        flutterProject.directory.childFile('.packages'),
+        logger: globals.logger,
+      );
+      final File pluginRegistrant = TizenProject.fromFlutter(flutterProject)
+          .managedDirectory
+          .childFile('generated_plugin_registrant.dart');
+
+      runnerDir = globals.fs.systemTempDirectory..createTempSync();
```
swift-kim, line 80: Did you mean
```suggestion
      runnerDir = globals.fs.systemTempDirectory.createTempSync();
```

lib/commands/test.dart
```diff
@@ -4,13 +4,118 @@
...
+  entrypoint.main();
+}
+''');
+        newArgs.add(runnerFile.absolute.path.toString());
+      }
+      await test.main(newArgs);
+    } finally {
+      if (runnerDir.existsSync()) {
+        runnerDir.deleteSync(recursive: true);
+      }
```
swift-kim, line 111: You don't have to delete temp directories manually.

lib/commands/test.dart
```diff
@@ -4,13 +4,118 @@
...
+
+import '${testFile.absolute.path.toString()}' as entrypoint;
+import '${pluginRegistrant.absolute.path.toString()}';
+
+Future<void> main() async {
+  registerPlugins();
+  entrypoint.main();
+}
+''');
+        newArgs.add(runnerFile.absolute.path.toString());
```
swift-kim, line 105: Redundant `toString()`'s.

swift-kim (commented):
Looks okay for now. I'm going to implement the third solution soon so you can use this workaround until then.

lib/commands/test.dart
swift-kim: Line 105 :)

swift-kim (approved):

PR #187, opened 2021-08-04
Do not chmod TPK contents recursively

### LINKED ISSUE #186: The run command hangs after building a TPK
This part

https://github.com/flutter-tizen/flutter-tizen/blob/f10b27565b7be44b37d31214ec30fa44dc5dbf7a/lib/tizen_tpk.dart#L83-L87

takes too long (up to 3 minutes) when launching an app with a large number of asset files (such as Gallery) on WSL1. The call to system's `chmod` takes 60-250 ms for each file and the Gallery app has around 1000 asset files.

```
...
[   +8 ms] Building a Tizen application in release mode... (completed in 14.0s)
[   +2 ms] ✓ Built build/tizen/tpk/io.flutter.demo.gallery-2.8.1.tpk (92.9MB).
[   +9 ms] executing: unzip -o -q /home/swift/Git/gallery/build/tizen/tpk/io.flutter.demo.gallery-2.8.1.tpk -d /tmp/flutter_tools.EZWRMG/RAJQQY
[+1385 ms] Exit code 0 from: unzip -o -q /home/swift/Git/gallery/build/tizen/tpk/io.flutter.demo.gallery-2.8.1.tpk -d /tmp/flutter_tools.EZWRMG/RAJQQY
[+152281 ms] Stopping app 'io.flutter.demo.gallery-2.8.1.tpk' on Tizen SM-R800.
[   +1 ms] executing: /home/swift/tizen-studio/tools/sdb -s 00000314f59bd733 shell app_launcher -l
...
```

swift-kim:
Fixes #186.

### DIFF
--- a/lib/tizen_tpk.dart
+++ b/lib/tizen_tpk.dart
@@ -80,10 +80,8 @@ class TizenTpk extends ApplicationPackage {
     // We have to manually restore permissions for files zipped by
     // build-task-tizen on Unix.
     // Issue: https://github.com/dotnet/runtime/issues/1548
-    await tempDir.list(recursive: true).forEach((FileSystemEntity entity) {
-      if (entity is File) {
-        globals.os.chmod(entity, '644');
-      }
+    tempDir.listSync().whereType<File>().forEach((File file) {
+      globals.os.chmod(file, '644');
     });
 
     final File manifestFile = tempDir.childFile('tizen-manifest.xml');


lib/tizen_tpk.dart
```diff
@@ -80,10 +80,8 @@ class TizenTpk extends ApplicationPackage {
     // We have to manually restore permissions for files zipped by
     // build-task-tizen on Unix.
     // Issue: https://github.com/dotnet/runtime/issues/1548
-    await tempDir.list(recursive: true).forEach((FileSystemEntity entity) {
-      if (entity is File) {
-        globals.os.chmod(entity, '644');
-      }
+    tempDir.listSync().whereType<File>().forEach((File file) {
+      globals.os.chmod(file, '644');
```
WonyoungChoi, line 84: @swift-kim Is the reason for unzipping and setting permissions to get the two XML files (`tizen-manifest.xml`, `author-signature.xml`) below?

If so, is the following code as possible?
```dart
globals.os.chmod(manifestFile, '644');
globals.os.chmod(signatureFile, '644');

```

swift-kim: Yes, it's also possible.

WonyoungChoi: How about making a utility function to extract a specific file from a zip file? I think we don't need to extract all files to tempdir. 

```dart
   File extractFromArchive(File archiveFile, String targetFilePath, {String mode, String tempPath});
```

swift-kim: The flutter-tizen tool is designed to be portable, and I'd like to only use functions available from `OperatingSystemUtils`. There's no such function in `OperatingSystemUtils` that extracts a single file from a zip file.

Also it makes the code less readable.

bwikbs: Is there an API to extract only specific file?

WonyoungChoi: Maybe we can use `package:archive` (https://pub.dev/packages/archive/example). `OperatingSystemUtils` is also using this package for Windows port. But I don't know about the performance of `package:archive`.

WonyoungChoi: It's not an important issue. We'd better improve it later if possible. :)

bwikbs: You're right! :smile: 
I think there is a no way to extract only a specific file... I just wonder if there's a way you've been thinking.. That's all !

WonyoungChoi: @bwikbs There is a way to extract a single file by using `package:archive`. this is my simple PoC, but it should be tested more  : 
```dart
import 'dart:io';
import 'package:archive/archive.dart';
import 'package:archive/archive_io.dart';

void main(List<String> arguments) {
  final extractedFile = extractFileFromArchive(
    'test.tpk',
    'tizen-manifest.xml',
    'out/tizen-manifest.xml',
  );

  print(extractedFile?.path);
}

File? extractFileFromArchive(
  String archivePath,
  String filePath,
  String outputPath,
) {
  final input = InputStream(File(archivePath).readAsBytesSync());
  final archive = ZipDecoder().decodeBuffer(input);
  for (final file in archive) {
    if (file.name == filePath) {
      return File(outputPath)
        ..createSync(recursive: true)
        ..writeAsBytesSync(file.content);
    }
  }
  return null;
}

```

bwikbs: @WonyoungChoi Thanks for the detailed example code! I'm Impressed 😭 .

And my original intention was not that this could be implemented or not. generally, extraction is not random access. I mean.. I thought that if I extracted two files, the same decoding would have to be repeated twice, so I thought this was a waste. 
If the problem is that there are unnecessary files in tmp path, it may be effective to unzip only once and delete unnecessary files... (Of course, your sample implementation is not, but I was thinking of the first interface)

WonyoungChoi (approved):

Issue #193, opened 2021-08-09
Rework the native plugin builder


swift-kim:
Currently many properties defined in plugins' `project_def.prop` are not supported (except `USER_LIBS`) because their source code is copied into a temp directory before build and all plugins referenced by an app are built together as a single shared lib project (due to https://github.com/flutter-tizen/flutter-tizen/pull/98).

My current plan is to split the build process into two stages:
1. Build each native plugin as a static lib (not a shared lib).
2. Build cpp_client_wrapper sources and link all static libs generated in 1. A single shared object is generated as a result.

I'm not sure whether this would be possible or not though.

PR #195, opened 2021-08-11
Rework the native plugin builder

swift-kim:
Implements #193.

- Each plugin is built as a static library, and included when the final shared object (`libflutter_plugins.so`) is built.
- Because the target project type is now `staticLib`, linker-specific properties defined in `project_def.prop` do not apply. However, you can still place your libs in `lib` or arch-specific directories under it, such as `lib/arm` and `lib/armel`. The libs will be automatically included during linking even though they are not defined anywhere.
  - `lib/arm/libstatic.a` precedes `lib/libstatic.a` if they both exist.

cc @WonyoungChoi

The inputs and outputs of the `tizen_native_plugins` target include:

(Inputs)
- `flutter-tizen/lib/build_targets/plugins.dart`
- `flutter-tizen/lib/tizen_sdk.dart`
- `APP_DIR/.packages` (to create a list of plugins to build)
- `APP_DIR/tizen/tizen-manifest.xml` (to check the target profile and API version)
- `flutter/bin/cache/artifacts/engine/tizen-common/cpp_client_wrapper/**` 
- `flutter/bin/cache/artifacts/engine/tizen-common/public/**`
- `flutter/bin/cache/artifacts/engine/tizen-ARCH-MODE/libflutter_tizen_PROFILE.so`
- `PLUGIN_DIR/tizen/project_def.prop`
- `PLUGIN_DIR/tizen/inc/**`
- `PLUGIN_DIR/tizen/src/**`
- `PLUGIN_DIR/tizen/lib/**`

(Outputs)
- `BUILD_DIR/tizen_plugins/lib/*.so` (copied from `PLUGIN_DIR/tizen/lib/`)
- `BUILD_DIR/tizen_plugins/include/*.h` (copied from `PLUGIN_DIR/tizen/inc/PLUGIN_CLASS.h`)
- `BUILD_DIR/tizen_plugins/libflutter_plugins.so` (build output)

### DIFF
--- a/lib/build_targets/plugins.dart
+++ b/lib/build_targets/plugins.dart
@@ -36,7 +36,7 @@ class NativePlugins extends Target {
 
   @override
   List<Source> get inputs => const <Source>[
-        Source.pattern('{FLUTTER_ROOT}/../lib/tizen_build_target.dart'),
+        Source.pattern('{FLUTTER_ROOT}/../lib/build_targets/plugins.dart'),
         Source.pattern('{PROJECT_DIR}/.packages'),
       ];
 
@@ -201,16 +201,20 @@ class NativePlugins extends Target {
       for (final Directory directory
           in pluginLibDirs.where((Directory d) => d.existsSync())) {
         for (final File lib in directory.listSync().whereType<File>()) {
-          if (lib.basename.endsWith('.so') || lib.basename.endsWith('.a')) {
+          final bool isSharedLib = lib.basename.endsWith('.so');
+          final bool isStaticLib = lib.basename.endsWith('.a');
+          if (isSharedLib || isStaticLib) {
             final String libName = getLibNameForFileName(lib.basename);
             if (userLibs.contains(libName)) {
               continue;
             }
             lib.copySync(libDir.childFile(lib.basename).path);
-            userLibs.add(getLibNameForFileName(lib.basename));
+            userLibs.add(libName);
 
             inputs.add(lib);
-            outputs.add(libDir.childFile(lib.basename));
+            if (isSharedLib) {
+              outputs.add(libDir.childFile(lib.basename));
+            }
           }
         }
       }
--- a/lib/build_targets/plugins.dart
+++ b/lib/build_targets/plugins.dart
@@ -34,6 +34,7 @@ class NativePlugins extends Target {
   @override
   List<Source> get inputs => const <Source>[
         Source.pattern('{FLUTTER_ROOT}/../lib/build_targets/plugins.dart'),
+        Source.pattern('{FLUTTER_ROOT}/../lib/tizen_sdk.dart'),
         Source.pattern('{PROJECT_DIR}/.packages'),
       ];
 
@@ -58,8 +59,8 @@ class NativePlugins extends Target {
         if (key == 'type') {
           if (value != 'staticLib') {
             logger.printStatus(
-              'The project type of ${plugin.name} plugin is $value,'
-              ' which is deprecated in favor of staticLib.',
+              'The project type of ${plugin.name} plugin is "$value",'
+              ' which is deprecated in favor of "staticLib".',
             );
           }
           line = 'type = staticLib';
@@ -147,13 +148,15 @@ class NativePlugins extends Target {
         pluginCopy.directory.path,
         configuration: buildConfig,
         arch: getTizenCliArch(buildInfo.targetArch),
-        predefine: '${buildInfo.deviceProfile.toUpperCase()}_PROFILE',
+        predefines: <String>[
+          '${buildInfo.deviceProfile.toUpperCase()}_PROFILE',
+        ],
         extraOptions: <String>[
           '-fPIC',
           '-I"${clientWrapperDir.childDirectory('include').path.toPosixPath()}"',
           '-I"${publicDir.path.toPosixPath()}"',
         ],
-        rootstrap: rootstrap,
+        rootstrap: rootstrap.id,
         environment: <String, String>{
           'PATH': getDefaultPathVariable(),
         },
@@ -266,7 +269,7 @@ USER_LIBS = ${userLibs.join(' ')}
         for (String className in pluginClasses)
           '-Wl,--undefined=${className}RegisterWithRegistrar',
       ],
-      rootstrap: rootstrap,
+      rootstrap: rootstrap.id,
       environment: <String, String>{
         'PATH': getDefaultPathVariable(),
       },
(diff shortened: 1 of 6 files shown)


templates/plugin/cpp/project_def.prop.tmpl
```diff
@@ -2,7 +2,7 @@
 # for details.
 
 APPNAME = {{projectName}}_plugin
-type = sharedLib
+type = staticLib
```
swift-kim, line 5: Existing plugins are not affected. The builder always uses `staticLib` as a project type when building plugins.

WonyoungChoi:
@swift-kim It feels like the build function is too long. How about separating the part that builds the static library into separate targets or private methods?

swift-kim:
@WonyoungChoi Splitting into multiple targets will make it hard to manage input and output dependencies.

Separating methods looks like a better option, but it will make the code much longer as many variables are shared between the two parts. They are closed related to each other.

WonyoungChoi:
@swift-kim Well.. I have a slightly different view. I think separating the target will make the input/output dependencies more clear. 
To put it in the pseudo Makefile grammar, 
```makefile
sharedlibrary: staticlibraryA, staticLibraryB
    // generating the shared library with the generated static libraries
staticLibraryA: headerfiles of packageA, sourcefiles of packageA
    // generating the static library with files of packageA
staticLibraryB: headerfiles of packageB, sourcefiles of packageB
    // generating the static library with files of packageB
```
The current implementation is that the main target uses all dependent source files as input.

swift-kim:
@WonyoungChoi That's a too broad view. In reality, there are more inputs and outputs than that, especially shared libs and static libs provided by each plugin.

lib/tizen_plugins.dart
```diff
@@ -66,6 +66,16 @@ class TizenPlugin extends PluginPlatform implements NativeOrDartPlugin {
   final String dartPluginClass;
   final String fileName;
 
+  TizenPlugin copyWith({Directory directory}) {
```
WonyoungChoi, line 69: How about changing to `copyTo` and including the directory copying. 
```dart
TizenPlugin copyTo(Directory destination) {
  if (destination == null) {
    throw Exception('destination is null');
  }
  copyDirectory(directory, destination);
  return TizenPlugin(
    ....
}
```

lib/build_targets/plugins.dart
```diff
@@ -64,151 +64,206 @@ class NativePlugins extends Target {
...
+      );
+      copyDirectory(plugin.directory, pluginCopy.directory);
+
+      final List<String> properties = <String>[];
+      for (String line in plugin.projectFile.readAsLinesSync()) {
+        if (line.startsWith('type =')) {
+          line = 'type = staticLib';
+        }
+        properties.add(line);
+      }
```
WonyoungChoi, line 130: maybe this part can be extracted to `_ensureStaticLibType(plugin)`. 
And I the type is not `staticLib`, I think the information that is has been changed should be notified as some log message.

lib/tizen_plugins.dart
swift-kim: Actually it was an adaptation of the "copyWith" pattern widely used when developing Flutter apps.

lib/build_targets/plugins.dart
swift-kim: Nice idea!

swift-kim:
I agree that the build method is complicated and hard to follow the overall flow at a glance. It's purposefully complicated, because a lot of input and output files involve during the build. I'll provide some map of inputs and outputs associated with the target if you need.

lib/build_targets/plugins.dart
```diff
@@ -64,151 +64,206 @@ class NativePlugins extends Target {
...
+        '-fPIC',
+        '-I"${clientWrapperDir.childDirectory('include').path.toPosixPath()}"',
+        '-I"${publicDir.path.toPosixPath()}"',
+        '-D${buildInfo.deviceProfile.toUpperCase()}_PROFILE',
+      ];
 
-      final Directory headerDir = plugin.directory.childDirectory('inc');
-      if (headerDir.existsSync()) {
-        headerDir
+      final RunResult result = await _processUtils.run(<String>[
```
WonyoungChoi, line 144: How about adding `buildNative()` method to `TizenSdk`?
```dart
Future<File> buildNative(
    TizenBuildInfo buildInfo,
    Rootstrap rootstrap, 
    Directory buildDirectory,
    String expectedOutputPath, { 
    List<String> extras,
    Map<String, String> environment,
    }) async {
      ....
```
this part can be more simple like below:
```dart
     final File libFile = await tizenSdk.buildNative(
        buildInfo,
        rootstrap,
        pluginCopy.directory,
        'lib$libName.a',
        extras: <String>[
          '-fPIC',
          '-I"${clientWrapperDir.childDirectory('include').path.toPosixPath()}"',
          '-I"${publicDir.path.toPosixPath()}"',
          '-D${buildInfo.deviceProfile.toUpperCase()}_PROFILE',
        ],
        environment: <String, String>{
          'PATH': getDefaultPathVariable(),
        },
      );
```

swift-kim: Thanks! I applied the suggestion but modified the method signature a bit to match the actual parameters of the `tizen build-native` command.

swift-kim: I forgot to apply this to the NativeTpk target in `package.dart`. I'll update it as well.

lib/tizen_sdk.dart
```diff
@@ -116,6 +122,36 @@ class TizenSdk {
...
+    String predefine,
+    List<String> extraOptions = const <String>[],
+    Rootstrap rootstrap,
+    Map<String, String> environment,
+  }) async {
+    assert(configuration != null);
+    assert(arch != null);
+
+    return _processUtils.run(<String>[
+      tizenSdk.tizenCli.path,
```
WonyoungChoi, line 139:
```suggestion
      tizenCli.path,
```

lib/tizen_sdk.dart
```diff
@@ -116,6 +122,36 @@ class TizenSdk {
...
+      'build-native',
+      '-C',
+      configuration,
+      '-a',
+      arch,
+      '-c',
+      compiler ?? defaultNativeCompiler,
+      if (predefine != null) ...<String>['-d', predefine],
```
WonyoungChoi, line 147: any cases where multiple predefine values come in?

swift-kim: I tested
```
tizen build-native -d "FIRST_DEFINE SECOND_DEFINE THIRD_DEFINE" ...
```
but it didn't work. The predefine option expects only one argument. If you have more than one user defines, they should go in `extraOptions`.

swift-kim:
```
tizen build-native -d FIRST_DEFINE -d SECOND_DEFINE -d THIRD_DEFINE ...
```
This worked. I'll make a change using `for`.

WonyoungChoi (approved):
LGTM 
👍 Thank you for your hard work. :)

bbrto21 (approved):

Issue #190, opened 2021-08-06
Document how to debug apps with VS Code


swift-kim:
https://github.com/sony/flutter-elinux/wiki/Debugging-with-VSCode

cc @xuelian-bai Did you know this? This is cool!

swift-kim:
I'm planning to let the tool automatically configure `launch.json` on app launch so that the user doesn't have to copy and paste `observatoryUri` manually. In that way, the user can just press the Run button (F5) to attach to a running app and start debugging. I'll prepare a PR soon.

cc @xuelian-bai

xuelian-bai:
That sounds good. 
I have one concern, in vscode, for example, run a java application, user can set breakpoint and run, user don't need to attach a running app manually, in this way, user can debug an app which may can't launch properly, but, if app can only be attached, the precondition is that the app can launch properly.
But, it seems that this method use offcial vscode flutter extension, which is not open source, I think maybe we have no other way to debug...

swift-kim:
@xuelian-bai Yes, I'm aware of that limitation. Only the "Attach and Debug" function can be supported in the way I mentioned. I'll try to find a way to enable "Run and Debug", but actually there could be no way to resolve the limitation. Still, I think the limitation is something that can be understanded by users as long as we document it properly.

> the precondition is that the app can launch properly.

FYI, there is an option `--start-paused` that can provided to the `flutter-tizen run` command. Code execution can be paused until a debugger is attached.

> offcial vscode flutter extension, which is not open source,

Why do you think it's not open-sourced: https://github.com/Dart-Code/Dart-Code :)

There's a chance to contribute to the project, as I mentioned as a second option in https://github.com/flutter-tizen/flutter-tizen/issues/88#issuecomment-823812365.

PR #196, opened 2021-08-13
Support attaching a debugger in VS Code

swift-kim:
Context: https://github.com/flutter-tizen/flutter-tizen/issues/190#issuecomment-897563757

Automatically creates or updates `.vscode/launch.json` file in the project directory to provide a debug configuration for the VS Code debugger. If the project is an example app of another plugin or module project, `.vscode/launch.json` in the parent directory is also created or updated.

Usage:
1. Run an app in either debug or profile mode using the `flutter-tizen run` command. (You can optionally specify the `--start-paused` option.)
2. Once the app is launched, open the **Run and Debug** menu in VS Code and choose **flutter-tizen: Attach (project_name)** from available configurations.
3. Press Run (▷) or F5 to attach to the running app.
4. Try breakpoints, widget inspector, etc.

### DIFF
--- a/lib/vscode_helper.dart
+++ b/lib/vscode_helper.dart
@@ -11,7 +11,7 @@ import 'package:flutter_tools/src/convert.dart';
 import 'package:flutter_tools/src/globals.dart' as globals;
 import 'package:flutter_tools/src/project.dart';
 
-const String kConfigNameAttach = 'Attach (flutter-tizen)';
+const String kConfigNameAttach = 'flutter-tizen: Attach';
 
 final RegExp _commentFormat =
     RegExp(r'(:?)(/\*([^*]|[\r\n]|(\*+([^*/]|[\r\n])))*\*+/|//.*\n?)');
@@ -28,14 +28,14 @@ String _removeComments(String jsonString) {
 
 void updateLaunchJsonFile(FlutterProject project, Uri observatoryUri) {
   Directory vscodeDir;
-  String cwd = r'${workspaceFolder}';
+  String workingDirectory = r'${workspaceFolder}';
   if (project.directory.basename == 'example') {
     final FlutterProject parentProject =
         FlutterProject.fromDirectory(project.directory.parent);
     if (parentProject.isPlugin || parentProject.isModule) {
       vscodeDir = parentProject.directory.childDirectory('.vscode')
         ..createSync(recursive: true);
-      cwd += '/example';
+      workingDirectory += '/example';
     }
   }
   vscodeDir ??= project.directory.childDirectory('.vscode')
@@ -74,7 +74,7 @@ void updateLaunchJsonFile(FlutterProject project, Uri observatoryUri) {
     if (config['name'] != kConfigNameAttach) {
       continue;
     }
-    config['cwd'] = cwd;
+    config['cwd'] = workingDirectory;
     config['observatoryUri'] = observatoryUri.toString();
   }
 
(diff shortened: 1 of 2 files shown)


lib/vscode_helper.dart
```diff
@@ -0,0 +1,83 @@
...
+  String cwd = r'${workspaceFolder}';
+  if (project.directory.basename == 'example') {
+    final FlutterProject parentProject =
+        FlutterProject.fromDirectory(project.directory.parent);
+    if (parentProject.isPlugin || parentProject.isModule) {
+      vscodeDir = parentProject.directory.childDirectory('.vscode')
+        ..createSync(recursive: true);
+      cwd += '/example';
+    }
+  }
```
WonyoungChoi, line 40: I think the `launch.json` file should be just located in  current project. Because users run and debug the example project itself, not a parent module or plugin. In general, users open the example project to debug it.

lib/vscode_helper.dart
```diff
@@ -0,0 +1,83 @@
...
+  vscodeDir ??= project.directory.childDirectory('.vscode')
+    ..createSync(recursive: true);
+
+  final File launchJsonFile = vscodeDir.childFile('launch.json');
+  String jsonString = '';
+  if (launchJsonFile.existsSync()) {
+    // Comments are allowed in launch.json file. Remove them to prevent
+    // decoding errors.
+    jsonString = _removeComments(launchJsonFile.readAsStringSync());
```
WonyoungChoi, line 49: just question :)  Is there any problem to parse comments with  `jsonDecoder` of `dart:convert` ?

swift-kim: Comments are not allowed in standard JSON files.

image

`jsonDecode` will throw an error when a comment is encountered.

WonyoungChoi: Oh.. I see. that's special thing about launch.json....  😲

swift-kim: Maybe I could create my own SAX-like JSON parser to keep the original comments, but it's not very easy :(

WonyoungChoi (approved):
LGTM 😃

swift-kim:
The difference between the URI logged to the console (Dart VM service URI) and the URI written to `launch.json` (DDS instance URI) comes from https://github.com/dart-lang/sdk/blob/master/pkg/dds/dds_protocol.md.

bbrto21:
I've always used Dev-tools for the web, it's so cool!!

bbrto21 (approved):

lib/vscode_helper.dart
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  Map<dynamic, dynamic> decoded = <dynamic, dynamic>{};
+  if (jsonString.isNotEmpty) {
+    try {
+      decoded = jsonDecode(jsonString) as Map<dynamic, dynamic>;
+    } on FormatException catch (error) {
+      globals.printError('Failed to parse ${launchJsonFile.path}:\n$error');
+      return;
```
xuelian-bai, line 54: If failed,why not just create a new one?

swift-kim: The file is not owned by the flutter-tizen tool but is managed by VS Code and the user, so we don't want to overwrite it arbitrarily.

Side note)
The main reason for the FormatException should be manually edited `launch.json` files which contain invalid "trailing commas". Similarly to "comments", trailing commas are not allowed in standard JSON but allowed in `launch.json` files which use extended JSON. I need to find a better way to handle this.

HakkyuKim (approved):
LGTM

Issue #143, opened 2021-06-21
Support multi package app


jkpu:
I need flutter  multi package app (UI app + service app).
Tizen native app and .NET app supports multi package. 
Each project can refer to other project's code and can access files and database in the package installed directory.
It will be very helpful if Tizen flutter app supports multi package.

Native app
https://docs.tizen.org/application/native/tutorials/process/app-dev-process/#developing-multiple-projects-as-a-combined-package

.NET app
https://developer.samsung.com/tizen/blog/en-us/2019/01/04/how-to-package-ui-and-service-applications-together-and-perform-them

pkosko:
I will check the feasibility of this feature if it is not assigned yet.

swift-kim:
@jkpu How does the TPK with UI app + service app look like? Is the source code of the two apps compiled into a single executable under `bin` or do they have separate entrypoints?

jkpu:
@swift-kim  No. UI app and service app has each own executable, they only package in same folder.
You can refer to native app example or .Net example in the above link.

following is the result I have tested with native UI app + service app 

```log
/opt/usr/home/owner/apps_rw/org.example.basicui
sh-3.2# ls -al
total 32
drwxr-xr-x   5 owner  users        4096 Jun 29 16:35 .
drwxr-x--- 106 owner  system_share 4096 Jun 29 16:35 ..
lrwxrwxrwx   1 app_fw app_fw         60 Jun 29 16:35 author-signature.xml -> /opt/usr/globalapps/org.example.basicui/author-signature.xml
lrwxrwxrwx   1 app_fw app_fw         43 Jun 29 16:35 bin -> /opt/usr/globalapps/org.example.basicui/bin
drwxr-sr-x   2 owner  system_share 4096 Jun 29 16:35 cache
drwxr-sr-x   2 owner  system_share 4096 Jun 29 16:35 data
lrwxrwxrwx   1 app_fw app_fw         43 Jun 29 16:35 lib -> /opt/usr/globalapps/org.example.basicui/lib
lrwxrwxrwx   1 app_fw app_fw         43 Jun 29 16:35 res -> /opt/usr/globalapps/org.example.basicui/res
drwxr-xr-x   3 owner  users        4096 Jun 29 16:35 shared
lrwxrwxrwx   1 app_fw app_fw         58 Jun 29 16:35 tizen-manifest.xml -> /opt/usr/globalapps/org.example.basicui/tizen-manifest.xml
[...]

PR #204, opened 2021-08-18
Set a proper version for the generated rootstrap definition

swift-kim:
Contributes to #143.

The version attribute of the generated rootstrap definition is used as part of the output TPK name when building a multi app. This change lets the attribute have a meaningful value.

```sh
# Before:
com.example.multi_app_flutter-null_armel_Debug.tpk

# After:
com.example.multi_app_Tizen-4.0_armel_Debug.tpk
```


bwikbs (approved):
I didn't understand where it is used... but it looks better than now.

bbrto21 (approved):

PR #205, opened 2021-08-18
Make Dart entrypoint setters public

swift-kim:
I think we can now provide stable APIs for setting Dart entry points.

The entry points can be set before the `Run()` method is called on app instances.

Example (C++):
```cpp
App app;
app.SetDartEntrypoint("anotherMain");
return app.Run(argc, argv);
```

Example (C#):
```csharp
var app = new App();
app.DartEntrypoint = "anotherMain";
app.Run(args);
```

Contributes to #143.

cc @pkosko


bwikbs (approved):
Cool!  👍

pkosko (approved):
Great! Much more elegant.

PR #221, opened 2021-08-30
Fix stop service

m-bialek:
I have changed some code responsible for stopping the service. Unfortunately, I could not find a way to call a function on app exit so I added a button.


swift-kim:
This shouldn't be fixed in this way. I suggested FFI as a temporary workaround for **your own project**. I didn't expect changing the app template itself.

PR #224, opened 2021-08-31
Add the daemon command

swift-kim:
- Enable the `daemon` command and the `--machine` option. The command is hidden by default and not intended for use by general users so I didn't add to `doc/commands.md`.
- Use `LoggerFactory` when creating a Logger instance.


HakkyuKim (approved):
LGTM

PR #227, opened 2021-09-02
Fix the example app detection logic

swift-kim:
Adding support for non-plugin packages that have example apps.

### DIFF
--- a/lib/vscode_helper.dart
+++ b/lib/vscode_helper.dart
@@ -29,7 +29,7 @@ void updateLaunchJsonFile(FlutterProject project, Uri observatoryUri) {
   if (project.directory.basename == 'example') {
     final FlutterProject parentProject =
         FlutterProject.fromDirectory(project.directory.parent);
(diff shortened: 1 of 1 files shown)


lib/vscode_helper.dart
```diff
@@ -29,7 +29,7 @@ void updateLaunchJsonFile(FlutterProject project, Uri observatoryUri) {
   if (project.directory.basename == 'example') {
     final FlutterProject parentProject =
         FlutterProject.fromDirectory(project.directory.parent);
-    if (parentProject.isPlugin || parentProject.isModule) {
+    if (parentProject.pubspecFile.existsSync() && parentProject.hasExampleApp) {
```
HakkyuKim, line 32: Isn't `parentProject.hasExampleApp` always true is this scope?

swift-kim: Oh, technically, yes.

HakkyuKim (approved):
LGTM

PR #228, opened 2021-09-03
Enable the simultaneous use of Dart plugins and custom Dart entrypoints

swift-kim:
Find entry points functions from the main dart file and inject to the generated main file.

Resolves the limitation mentioned in https://github.com/flutter-tizen/flutter-tizen/pull/171 ("The use of custom entrypoints may break ...").

Note that this change doesn't take account of all possible uses of the `vm:entry-point` pragma described in https://github.com/dart-lang/sdk/blob/master/runtime/docs/compiler/aot/entry_point_pragma.md#syntax. Only functions are allowed and the second parameter of `@pragma` is not honored.

@jkpu @pkosko

See also: The AST analyzer (https://github.com/dart-lang/sdk/blob/master/pkg/analyzer/doc/tutorial/ast.md)

### DIFF
--- a/lib/tizen_plugins.dart
+++ b/lib/tizen_plugins.dart
@@ -5,6 +5,10 @@
 
 // @dart = 2.8
 
+import 'package:analyzer/dart/analysis/analysis_context.dart';
+import 'package:analyzer/dart/analysis/analysis_context_collection.dart';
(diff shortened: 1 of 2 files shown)


bbrto21 (approved):
I tested using Dart plugin, it works fine.
LGTM

Issue #223, opened 2021-08-31
Add test code


swift-kim:
I previously tried to add some unit tests for our tool code but gave up soon because the code was not mature enough then and it was obviously not worth taking the efforts. Now I'd like to add some hermetic/permeable tests at least for some commands like `create` and `build` that are easy to break.

HakkyuKim:
I'd like solve this if no one has started yet. 

@swift-kim 
Are you thinking of implementing `build_apk_test.dart` and `create_test.dart` for Tizen projects?

swift-kim:
@HakkyuKim Generally, yes. The unit of testing target could be either function (unit testing - `general.shard`) or command (`commands.shard`).

The important principle I have in my mind is that we need to test only "real" things that are meaningful for us. Testing trivial scenarios or just increasing the test coverage of the code is not my goal. I don't plan to "copy" tests from flutter_tools (although you can refer to how they work), and I want to test only Tizen-specific things that are not tested by the original flutter_tools.

I had some specific test cases that I wanted to test, and I'll list them up (or implement) later. We need to set up the testing environment first. When I tried last time I needed to copy the entire `test` directory from flutter_tools and make [this change](https://github.com/swift-kim/flutter-tizen/commit/f25aa797383565ff72f812869932f627d1846d70#diff-06d52610e4161af0c5f25ee28845eee6e809585eb8cee475dc7b3d68b9345d96). It would be better if there's another way that doesn't require copying the directory.

HakkyuKim:
@swift-kim 
I see, currently I can think of service app and multi-app for Tizen specific tests, I'm sure there are more. It would be helpful to have some scenarios to start with.

Regarding the testing environment, I can't figure out the best way to depend on `test/src` because I'm unsure about the factors involved. Is the most concerning part resolving future changes to the `test/src` directory?

Currently I see two options:
- As you mentioned, copying the `test/src` directory and make changes for flutter-tizen, or
- Creating single file `test/tizen_common.dart` export all modules in `test/src` but hide the ones that we need to implement for flutter-tizen.

This is another question, is there anything I should be aware of if I use the import statement, `import ../../flutter/packages/flutter_tools/test/src/context.dart;` from say `test/tizen_create_test.dart`?

swift-kim:
[...]

PR #231, opened 2021-09-10
Setup test environment

HakkyuKim:
Contributes to #223.

This PR adds files that imports essential functions(`testUsingContext` or `testWithoutContext`) for writing tests. `testUsingContext` is used to test classes with context-based injection and `testWithoutContext` is used to write tests for classes that don't use contexts.

To run tests, you first have to set an environment variable `FLUTTER_ROOT` to the root of flutter repository. Running the following command from `flutter-tizen` repository will run tests:
```
FLUTTER_ROOT=flutter flutter-tizen test
```


test/create_test.dart
```diff
@@ -0,0 +1,117 @@
+// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
```
swift-kim, line 1: Please remove this file. I'll write a test regarding the multi app creation scenario later.

HakkyuKim: Okay, removed.

swift-kim (approved):

Issue #210, opened 2021-08-21
Rework the app_control channel in the embedder


swift-kim:
Related to https://github.com/flutter-tizen/flutter-tizen/pull/203#discussion_r690897502.

The below is just my current plan and subject to change.

---

### 1) Re-design the channel APIs.

The current reply channel is not quite necessary.

- As-is: Three channels (`app_control_method`, `app_control_event`, `app_control_reply`)
- To-be: Two channels (`app_control`, `app_control_event`)

**Scenario a) Send a request without waiting.**

As-is (pseudo code):
```dart
int id = await methodChannel.create();
await methodChannel.setAppControlData(id: id, appId: 'com.example.appid');
await methodChannel.sendLaunchRequest(id: id);
```

To-be:
```dart
Map request = {appId: 'com.example.appid'};
await methodChannel.sendLaunchRequest(request: request);
```

**Scenario b) Send a request and wait for a reply.**

As-is:
```dart
int id = await methodChannel.create();
await methodChannel.setAppControlData(id: id, appId: 'com.example.appid');
await methodChannel.sendLaunchRequest(id: id, waitForReply: true);

replyChannel.listen((Map response) {
  int id = response['id']; // The request id.
  Map reply = response['reply']; // The reply app_control data. reply['id'] has no meaning.
  String result = response['result'];
});
```

To-be:
```dart
Map request = {appId: 'com.example.appid'};
Map response = await methodChannel.sendLaunchRequest(request : request, waitForReply: true);
Map reply = response['reply'];
String result = response['result'];
```

**Scenario c) Receive a request without replying.**

As-is:
```dart
eventChannel.listen((Map received)) {
  String callerAppId = received['callerId'];
  String appId = received['appId'];
  String operation = received['operation'];
  ...
});
```

To-be:
```dart
eventChannel.listen((Map received)) {
  bool shouldReply = received['shouldReply']; // app_control_is_reply_requested()
  String callerAppId = received['callerAppId'];
  String appId = received['appId'];
  String operation = received['operation'];
  ...
});
```

**Scenario d) Receive a request to reply.**

As-is:
```dart
eventChannel.listen((Map received)) {
  int requestId = received['id'];
  methodChannel.reply(requestId: requestId, result: 'appStarted');
});
```

To-be:
```dart
eventChannel.listen((Map received)) {
  bool shouldReply = received['shouldReply'];
  if (shouldReply) {
    Map reply = {};
[...]

PR #235, opened 2021-09-17
Update the multi app template

swift-kim:
Replace the multi app template with a simplified version of [`tizen_app_control`'s example app](https://github.com/flutter-tizen/plugins/blob/master/packages/tizen_app_control/example/lib/main.dart).

`tizen_app_control` should be published first to be able to use this template. You can modify `pubspec.yaml` as follows for testing until then.

```yaml
tizen_app_control:
  git:
    url: https://github.com/flutter-tizen/plugins
    path: packages/tizen_app_control
```

Closes https://github.com/flutter-tizen/flutter-tizen/issues/210.

### DIFF
--- a/lib/commands/create.dart
+++ b/lib/commands/create.dart
@@ -53,11 +53,14 @@ class TizenCreateCommand extends CreateCommand {
       return super.runCommand();
     }
 
-    // Check if the main.dart file exists before running super.runCommand().
+    // Check if main.dart/pubspec.yaml exists before running super.runCommand().
     final File mainFile =
         projectDir.childDirectory('lib').childFile('main.dart');
+    final File pubspecFile = projectDir.childFile('pubspec.yaml');
     final bool overwriteMainFile =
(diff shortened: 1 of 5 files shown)


lib/commands/create.dart
```diff
@@ -127,6 +131,24 @@ class TizenCreateCommand extends CreateCommand {
...
+        );
+        pubspecFile.writeAsStringSync(pubspec.join('\n') + '\n');
+
+        await pub.get(
+          context: PubContext.create,
+          directory: projectDir.path,
+          offline: boolArg('offline'),
+          generateSyntheticPackage: false,
+        );
+      }
```
swift-kim, line 151: This is a temporary solution and I'm preparing another patch that may remove the entire if statement between line 127-153. I'd like to more focus on the template change (in `main.dart.tmpl`) in this PR.

HakkyuKim (approved):
LGTM

PR #236, opened 2021-09-22
Remove temporary workarounds in create.dart

swift-kim:
- Support the `flutter-tizen create --platforms` option. (Now `--platforms tizen` must be explicitly provided when creating a plugin project.)
- Do not copy `templates/template_manifest.json` which is not necessary any more.
- Add tests.


HakkyuKim (approved):
LGTM

PR #248, opened 2021-10-15
Update commands.md

swift-kim:
- Add the `custom-devices` command.
- Refine existing words.


swift-kim:
@HakkyuKim I actually omitted the information not to confuse users between the `drive` and `test` commands, but I'll state it more clearly in the doc.

PR #263, opened 2021-10-29
Ignore avoid_private_typedef_functions in generated_main.dart

swift-kim:
A lint warning is generated because `_NullaryFunction` is referenced only once unless multiple entry points exist.


WonyoungChoi (approved):
LGTM

Issue #127, opened 2021-05-28
Platform channel implementation in C#


swift-kim:
- [x] PoC for Dart to C# messaging
- Implement Codec and Channel types
  - [x] BasicMessageChannel, StringCodec (#253)
  - [x] StandardMessageCodec (#266)
  - [x] MethodChannel, StandardMethodCodec (#267)
  - [x] EventChannel (#270)
- [ ] Add a tool support for building C# plugins - maybe?

Note that only primitive types can be supported, similarly to other language implementations and Dart isolates.

swift-kim:
TV API) For example, the Tizen.TV.Multimedia.ESPlayer class mentioned in https://github.com/flutter-tizen/plugins/issues/44 is defined in

- https://developer.samsung.com/smarttv/develop/api-references/tizenfx-tv-api-references/tizen.tv.multimedia/esplayer-class.html (public API references)
- https://platz.sec.samsung.net/cs/item/127238465 (internal API references)

@haesik @bbrto21 @bwikbs FYI

swift-kim:
The following is the example code that I used for PoC.

Dart (`lib/main.dart`):
```dart
import 'package:flutter/services.dart';

void main() {
  runApp(MyApp());

  const BasicMessageChannel channel =
      const BasicMessageChannel('test_message_channel', StringCodec());

  channel.setMessageHandler((dynamic message) async {
    print('Received message (Dart): $message (${message.runtimeType})');

    channel.send('Hello from Dart! 하이!');
  });
}
```

C# (`tizen/App.cs`):
```csharp
using System;
using System.Runtime.InteropServices;
using System.Text;
using static Tizen.Flutter.Embedding.Interop;

...

protected override void OnResume()
{
    const string channelName = "test_message_channel";

    var messenger = FlutterDesktopEngineGetMessenger(Handle);
    var sentMessage = Encoding.UTF8.GetBytes("Hello from C#! 你好!");
    var sentMessageHandle = GCHandle.Alloc(sentMessage, GCHandleType.Pinned);
    FlutterDesktopMessengerSend(messenger, channelName, sentMessageHandle.AddrOfPinnedObject(), (uint)sentMessage.Length + 1);
    sentMessageHandle.Free();

    FlutterDesktopMessengerSetCallback(messenger, channelName, (FlutterDesktopMessenger _, IntPtr received, IntPtr user_data) =>
    {
[...]

PR #267, opened 2021-11-03
embedding/csharp: Add MethodChannel and related classes

WonyoungChoi:
- #127
- Add `MethodChannel`, `StandardMethodCodec` and related classes
- Add tests for the added classes

#### Usage
```c#
            var channel = new MethodChannel("test_method_channel");
            channel.SetMethodCallHandler(async (call) =>
            {
                if (call.Method == "method1")
                {
                    // handle the "method1" method and return a result
                    return new ArrayList() { 1, "TEST_STRING", true };
                }
                return null;
            });
```

### DIFF
--- a/embedding/csharp/Tizen.Flutter.Embedding.Tests/Channels/MethodChannelTests.cs
+++ b/embedding/csharp/Tizen.Flutter.Embedding.Tests/Channels/MethodChannelTests.cs
@@ -155,6 +155,49 @@ namespace Tizen.Flutter.Embedding.Tests.Channels
 
                 messenger.Received().SetMessageHandler(Arg.Is<string>(x => x == TEST_CHANNEL_NAME), null);
             }
+
+            [Fact]
+            public void Ensures_Error_Response_When_FlutterException_Is_Thrown()
+            {
+                var messenger = Substitute.For<IBinaryMessenger>();
+                var codec = StandardMethodCodec.Instance;
+                var channel = new MethodChannel(TEST_CHANNEL_NAME, StandardMethodCodec.Instance, messenger);
+                byte[] testCall = codec.EncodeMethodCall(new MethodCall("test", 1));
+                byte[] expected = codec.EncodeErrorEnvelope("E0001", "TEST_MESSAGE", "TEST_DETAIL");
+                messenger.When(x => x.SetMessageHandler(TEST_CHANNEL_NAME, Arg.Any<BinaryMessageHandler>()))
+                         .Do(async x =>
+                         {
+                             var binaryHandler = x[1] as BinaryMessageHandler;
+                             byte[] result = await binaryHandler(testCall);
+                             for (int i = 0; i < expected.Length; i++)
+                             {
+                                 Assert.Equal(expected[i], result[i]);
+                             }
+                         });
+                channel.SetMethodCallHandler((call) =>
+                {
+                    throw new FlutterException("E0001", "TEST_MESSAGE", "TEST_DETAIL");
+                });
+            }
+
+            [Fact]
+            public void Ensures_Empty_Response_When_MissingPluginException_Is_Thrown()
+            {
+                var messenger = Substitute.For<IBinaryMessenger>();
+                var channel = new MethodChannel(TEST_CHANNEL_NAME, StandardMethodCodec.Instance, messenger);
+                byte[] testCall = StandardMethodCodec.Instance.EncodeMethodCall(new MethodCall("test", 1));
+                messenger.When(x => x.SetMessageHandler(TEST_CHANNEL_NAME, Arg.Any<BinaryMessageHandler>()))
+                         .Do(async x =>
+                         {
+                             var binaryHandler = x[1] as BinaryMessageHandler;
+                             byte[] result = await binaryHandler(testCall);
+                             Assert.Null(result);
+                         });
+                channel.SetMethodCallHandler((call) =>
+                {
+                    throw new MissingPluginException("Not found method.");
+                });
+            }
         }
     }
 }
--- a/embedding/csharp/Tizen.Flutter.Embedding.Tests/Channels/MethodChannelTests.cs
+++ b/embedding/csharp/Tizen.Flutter.Embedding.Tests/Channels/MethodChannelTests.cs
@@ -1,0 +1,160 @@
+﻿// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+using System;
+using System.Linq;
+using NSubstitute;
+using Xunit;
+
+namespace Tizen.Flutter.Embedding.Tests.Channels
+{
+    public class MethodChannelTests
+    {
+        private const string TEST_CHANNEL_NAME = "TEST_METHOD_CHANNEL";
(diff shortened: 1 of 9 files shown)


embedding/csharp/Tizen.Flutter.Embedding/Channels/MethodChannel.cs
```diff
@@ -0,0 +1,156 @@
...
+                    return Codec.EncodeSuccessEnvelope(await handler(call));
+                }
+                catch (FlutterException fe)
+                {
+                    return Codec.EncodeErrorEnvelope(fe.Code, fe.Message, fe.Details, fe.StackTrace);
+                }
+                catch (Exception e)
+                {
+                    return Codec.EncodeErrorEnvelope("error", e.Message, null);
+                }
```
swift-kim, line 151: Shouldn't this be like
```suggestion
                catch (FlutterException fe)
                {
                    return Codec.EncodeErrorEnvelope(fe.Code, fe.Message, fe.Details);
                }
                catch (Exception e)
                {
                    return Codec.EncodeErrorEnvelope("error", e.Message, null, e.StackTrace);
                }
```
Also if the user wants to indicate that a **certain** method is **NotImplemented**, should they set the method call handler of the **entire** channel to null? Is there something that corresponds to Dart's [MissingPluginException](https://api.flutter.dev/flutter/services/MissingPluginException-class.html) or Java's [MethodChannel.Result.notImplemented](https://api.flutter.dev/javadoc/io/flutter/plugin/common/MethodChannel.Result.html)?

Let's say there's a simple C++ plugin that only implements the `getPlatformVersion` method:

```cpp
channel->SetMethodCallHandler(
    [plugin = this](const auto &call, auto result) {
      plugin->HandleMethodCall(call, std::move(result));
    });

void HandleMethodCall(
    const flutter::MethodCall<flutter::EncodableValue> &method_call,
    std::unique_ptr<flutter::MethodResult<flutter::EncodableValue>> result) {
  if (method_call.method_name().compare("getPlatformVersion") == 0) {
    result->Success(flutter::EncodableValue("1.0"));
  } else {
    result->NotImplemented();
  }
}
```
Then how would you re-implement this plugin in C#?

embedding/csharp/Tizen.Flutter.Embedding/Channels/StandardMethodCodec.cs
```diff
@@ -0,0 +1,141 @@
...
+                    MessageCodec.WriteValue(writer, exception.StackTrace);
+                }
+                else
+                {
+                    MessageCodec.WriteValue(writer, errorDetails);
+                }
+                if (!string.IsNullOrEmpty(errorStacktrace))
+                {
+                    MessageCodec.WriteValue(writer, errorStacktrace);
+                }
```
swift-kim, line 106: Is it okay if `errorDetails` is null? As far as I can see, the only possible cases include
- both `errorDetails` and `errorStacktrace` are null
- `errorDetails` is null and `errorStacktrace` is non-null
- `errorDetails` is non-null and `errorStacktrace` is null

embedding/csharp/Tizen.Flutter.Embedding/Channels/MethodChannel.cs
WonyoungChoi: I missed the part of `NotImplemented()`. I will consider adding the `Result` interface.

embedding/csharp/Tizen.Flutter.Embedding/Channels/StandardMethodCodec.cs
WonyoungChoi: It's okay. If the `errorDetails` is null, the zero(`0`) value as a `NULL` identifier will be written well.

embedding/csharp/Tizen.Flutter.Embedding/Channels/MethodChannel.cs
WonyoungChoi: I added `MissingPluginException` instead of `Result` interface because Java style implementation is not appropriate when using the async function.

WonyoungChoi: #### example:
```c#
            channel.SetMethodCallHandler(async (call) =>
            {
                switch (call.Method)
                {
                    case "method1":
                        return new ArrayList() { 1, "TEST_STRING", true };
                    case "method2":
                        throw new FlutterException("my001", "test error!!", null);
                    default:
                        throw new MissingPluginException();
                }
            });
```

WonyoungChoi: Actually, we can use other system exceptions instead of adding new `MissingPluginException`. for example `System.MissingMethodException`,  `System.NotSupportedException` or `System.NotImplementedException`. @swift-kim how about your opinion?

swift-kim: `MissingPluginException` feels slightly better for me for consistency with the Dart side implementation. (You could also replace `FlutterException` with `PlatformException` if you want.)

WonyoungChoi: > for consistency with the Dart side 
I agree to be consistent with Dart implementation. In this sense, `PlatformException` would be better. thanks.

WonyoungChoi: > I agree to be consistent with Dart implementation. ~In this sense, `PlatformException` would be better. thanks.~
Sorry, I would like to keep `FlutterException` again because it means that Platform and Flutter have opposite relationships.
* PlatformException : Thrown to indicate that a platform interaction failed in the platform plugin
* FlutterException: Thrown to indicate that a Flutter method invocation failed on the Flutter side.
In fact, `InvokeMethod` and `InvokeMethodAsync` of `MethodChannel` can throw this exception when an error response is received.

swift-kim: Ah, I thought they were referring to the same thing. Thanks for the explanation.

embedding/csharp/Tizen.Flutter.Embedding.Tests/Channels/MethodChannelTests.cs
```diff
@@ -0,0 +1,203 @@
...
+                messenger.Received().SetMessageHandler(Arg.Is<string>(x => x == TEST_CHANNEL_NAME), null);
+            }
+
+            [Fact]
+            public void Ensures_Error_Response_When_FlutterException_Is_Thrown()
+            {
+                var messenger = Substitute.For<IBinaryMessenger>();
+                var codec = StandardMethodCodec.Instance;
+                var channel = new MethodChannel(TEST_CHANNEL_NAME, StandardMethodCodec.Instance, messenger);
+                byte[] testCall = codec.EncodeMethodCall(new MethodCall("test", 1));
```
swift-kim, line 165: Maybe
```suggestion
                byte[] testCall = codec.EncodeMethodCall(new MethodCall(TEST_METHOD_NAME, 1));
```
(also in line 188)

swift-kim (commented):
Otherwise overall looks good to me.

swift-kim (approved):

Issue #260, opened 2021-10-28
Cannot attach dev tool to physical device


lenchis001:
Hello, everyone. I have a problem with attaching of dev tool when I run a test app to physical device.

What I did:
1. I created an app using "flutter-tizen create test_app";
2. I launched TV 6.0 emulator;
3. I run "flutter-tizen run lib/main.dart".

I see that url to dev tool is available:

<img width="1440" alt="Screenshot 2021-10-28 at 09 53 44" src="https://user-images.githubusercontent.com/15190316/139203015-ba951848-6d38-466d-ad9d-61155082f9bc.png">

Then I stoped the emulator and connected to a physical device and made "flutter-tizen run lib/main.dart" again:

![image](https://user-images.githubusercontent.com/15190316/139204668-1b7eadaa-61d1-48bd-8ba4-df05666de768.png)

There are no any other output in terminal. The application has started but url to dev tool is not available.

So the question is: Is it a known issue?
May be there is a workaround/fix for that?

Thanks.

swift-kim:
Okay, now I see the Web SDK uses the Web Inspector (which is based on Chrome Developer Tools) as a debugging tool and all logs are available in the "Console" tab. I think I'll have to re-implement the entire logging system based on websocket or something without relying on Tizen's dlog.

Until then, you can use the following trick to connect to your running app and start debugging.

1. Add the following code to somewhere in your app's widget tree, in order to display the app's observatory URI once it's launched.<p>
   ```dart
   import 'dart:developer';
   
   String observatoryUri = '';

   @override
   void initState() {
     super.initState();

     Service.getInfo().then((serviceInfo) {
       setState(() {
         observatoryUri = serviceInfo.serverUri.toString();
       });
     });
   }

   @override
   Widget build(BuildContext context) {
     ...
     Text('ObservatoryUri: $observatoryUri'),
     ...
   }
   ```
2. Launch the app and you will see a URI that looks like:<p>
   ```sh
   http://127.0.0.1:12345/iVJVuQtd5eQ=/
   ```
3. Connect to the VM in either of the following ways.
   - Manually launch Flutter DevTools from the command line. Please see https://flutter.dev/docs/development/tools/devtools/cli for how to start DevTools.
   - If you're using VS Code, import the app project into your workspace and create a file named `.vscode/launch.json` (modify if it already exists). Then, fill the file with the following contents.<p>
     ```json
     {
         "version": "0.2.0",
         "configurations": [
             {
[...]

PR #271, opened 2021-11-08
Replace TizenDlogReader with ForwardingLogReader

swift-kim:
Potentially fixes https://github.com/flutter-tizen/flutter-tizen/issues/260.

- Create `ForwardingLogReader` according to the embedder side change https://github.com/flutter-tizen/engine/pull/203 and remove `TizenDlogReader`. The tool doesn't rely on the device dlog any more and uses the forwarding logger only.
- Remove trailing newlines from engine arguments (`flutter_app.cc`).
- Unsupport commands `attach` and `logs`. (The logging port can be only passed on app launch.)

Depends on https://github.com/flutter-tizen/engine/pull/203.


lib/forwarding_log_reader.dart
```diff
@@ -0,0 +1,148 @@
...
+          if (line.isEmpty ||
+              _filteredTexts.any((String text) => line.contains(text))) {
+            continue;
+          }
+          _linesController.add(_colorizePrefix(line));
+        }
+      },
+      onError: (Object error) {
+        globals.printError(error.toString());
+      },
```
HakkyuKim, line 105: I'm not entirely sure about this because I couldn't test on sockets(`addError`  is blocked for sockets). Don't we have to do the same here as `onDone`?

swift-kim: I was not sure if I should simply ignore the error and let the socket continue reading data or regard it as a fatal error. I'll handle it more aggressively.

HakkyuKim (approved):
LGTM!

swift-kim:
Rebased upon master.

Issue #277, opened 2021-11-15
Problems with the order of libs used by plugin build


bwikbs:
In the following PR, message-loop lib used by lwe was added.
https://github.com/flutter-tizen/plugins/pull/246

And especially when building arm, I get the following error.
https://github.com/flutter-tizen/plugins/runs/4186671585?check_suite_focus=true

I'm guessing this is a matter of the ordering of the libraries used when  linking.

And.. this is workaround for better understanding
[lib/build_targets/plugins.dart](https://github.com/bwikbs/flutter-tizen/commit/f32bade87f17d102ebcece62b2c85a18852ebccc)

swift-kim:
We might need to give the `-pthread` option to the native compiler.

swift-kim:
Oh, the `-lpthread` option is already present in the build command (provided by Tizen SDK's `makefile.mk`), and it really looks like a matter of linking order as you said.

bwikbs:
You're right. And for now, I don't have any idea except workaround. Any suggestion?

swift-kim:
Here is my fix which is not very different from your solution: https://github.com/flutter-tizen/flutter-tizen/pull/278

bwikbs:
Wow! It's very fast :+1: 
Of course I love it!. My only concern is  you!! :smile:

PR #278, opened 2021-11-15
Let -lpthread come before any other linking options

swift-kim:
Fixes https://github.com/flutter-tizen/flutter-tizen/issues/277.


bwikbs (approved):
:100: Thanks!!

PR #279, opened 2021-11-19
Replace git restore with git checkout

swift-kim:
The `git restore` command requires Git 2.23 and is not generally available on some old Linux distributions.

Reported by @haesik.


haesik (approved):
Thanks for the quick update.

PR #289, opened 2021-12-14
Unsupport the custom-devices command

swift-kim:
The command is not relevant for Tizen app development.

### DIFF
--- a/lib/executable.dart
+++ b/lib/executable.dart
@@ -18,7 +18,6 @@ import 'package:flutter_tools/src/base/template.dart';
 import 'package:flutter_tools/src/cache.dart';
 import 'package:flutter_tools/src/commands/analyze.dart';
 import 'package:flutter_tools/src/commands/config.dart';
-import 'package:flutter_tools/src/commands/custom_devices.dart';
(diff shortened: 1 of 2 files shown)


lib/executable.dart
```diff
@@ -103,16 +102,6 @@ Future<void> main(List<String> args) async {
...
-        customDevicesConfig: globals.customDevicesConfig,
-        operatingSystemUtils: globals.os,
-        terminal: globals.terminal,
-        platform: globals.platform,
-        featureFlags: featureFlags,
-        processManager: globals.processManager,
-        fileSystem: globals.fs,
-        logger: globals.logger,
-      ),
       DaemonCommand(hidden: !verboseHelp),
```
HakkyuKim, line 105: Can we also remove `DaemonCommand` if it's not supported?

swift-kim: If I remember correctly, the daemon command worked without issue when I tested before. The command is hidden by default so it's not listed in the document.

HakkyuKim: Ah I see, I asked because it was listed as an unsupported command.

HakkyuKim (approved):

Issue #286, opened 2021-12-10
Migrate to Flutter 2.8


swift-kim:
https://medium.com/flutter/announcing-flutter-2-8-31d2cb7e19f5

PR #296, opened 2021-12-16
Implement TizenArtifacts and unsupport AZURE_BUILD_ID

swift-kim:
Part of https://github.com/flutter-tizen/flutter-tizen/issues/286 and https://github.com/flutter-tizen/engine/issues/224.

Reland `TizenArtifacts` which has previously been removed by https://github.com/flutter-tizen/flutter-tizen/pull/150, to enable overriding  `gen_snapshot` paths for Tizen during app build.

Notes:
- Only Linux `gen_snapshot` artifacts are currently included in the [Tizen engine 2.8.0 (1) pre-release](https://github.com/flutter-tizen/engine/releases/tag/790d2b0). I'll update the artifacts when https://github.com/flutter-tizen/engine/issues/224 is implemented by @WonyoungChoi. If you are on Windows or macOS, you can manually build `gen_snapshot` and copy it to the cached artifact directory corresponding to your host platform (e.g. `flutter/bin/cache/artifacts/engine/tizen-arm-release/windows-x64`).
- If anyone is using **flutter-tizen** to build **Android** apps, beware that this change may break Android release and profile builds. This is inevitable because we use the same `TargetPlatform` names for Android and Tizen.


lib/tizen_artifacts.dart
```diff
@@ -0,0 +1,70 @@
...
+      assert(mode != null, 'Need to specify a build mode.');
+      assert(mode != BuildMode.debug,
+          'Artifact $artifact only available in non-debug mode.');
+      final String arch =
+          _getArchForTargetPlatform(platform ?? TargetPlatform.android_arm);
+      final String hostPlatform =
+          getNameForHostPlatform(getCurrentHostPlatform());
+      return _getEngineArtifactsDirectory(arch, mode!)
+          .childDirectory(hostPlatform)
+          .childFile('gen_snapshot')
```
WonyoungChoi, line 53: Do we need ".exe" suffix on windows?

lib/tizen_artifacts.dart
```diff
@@ -0,0 +1,70 @@
...
+    EnvironmentType? environmentType,
+  }) {
+    if (artifact == Artifact.genSnapshot) {
+      assert(mode != null, 'Need to specify a build mode.');
+      assert(mode != BuildMode.debug,
+          'Artifact $artifact only available in non-debug mode.');
+      final String arch =
+          _getArchForTargetPlatform(platform ?? TargetPlatform.android_arm);
+      final String hostPlatform =
+          getNameForHostPlatform(getCurrentHostPlatform());
```
WonyoungChoi, line 50: We need to check the `supported host platform` here. `linux-arm64`, `darwin-arm` are out of our attention now. `darwin-arm` may be need to support `Apple Silicon` later.

WonyoungChoi, line 50: We need to check the `supported host platform` here. we don't support `linux-arm64` now.

WonyoungChoi:
Now that `tizen-arm-release.zip` contains the `gen_snapshot` binary, but when the CI works are finished, the artifacts will be separated to download only the artifact required for the host platform.

lib/tizen_artifacts.dart
swift-kim: I believe we don't. I'll double check later when gen_snapshots for Windows become available.

lib/tizen_artifacts.dart
swift-kim: I added an assert although it's not very meaningful because asserts are not enabled for `dart run` by default.

PR #306, opened 2021-12-30
Add isDotnet property to TizenTpk

swift-kim:
- Implement `TizenManifest.applicationType` and `TizenTpk.isDotnet`. The properties are not currently used but is preparation for https://github.com/flutter-tizen/plugins/issues/295.
- Refactor: Use `late final` (new in Dart 2.12) to avoid evaluating the same property multiple times.

### DIFF
--- a/lib/tizen_tpk.dart
+++ b/lib/tizen_tpk.dart
@@ -78,8 +78,11 @@ class TizenTpk extends ApplicationPackage {
   /// The SHA512 signature.
   final Signature? signature;
 
-  /// The application id if applicable.
+  /// The application ID.
   String get applicationId => manifest.applicationId;
+
+  /// Whether the application is a .NET application.
+  bool get isDotnet => manifest.applicationType == 'dotnet';
 
   @override
   String get name => file.basename;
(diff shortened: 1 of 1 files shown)


lib/tizen_tpk.dart
```diff
@@ -144,41 +147,43 @@ class TizenManifest {
...
-        globals.printTrace(
-            'tizen-manifest.xml: <$tag> is not officially supported.');
-      }
-      _applicationId = application.getAttribute('appid');
-      if (applications.length > 1) {
-        globals.printTrace(
-            'tizen-manifest.xml: Found ${applications.length} application declarations. '
-            'Using the first one: <$tag appid="$_applicationId">');
-      }
+  late final Iterable<XmlElement> _applications = () {
```
WonyoungChoi, line 156: If you're going to use only the first element of applications anyway, why do you have to take the list of applications?
Are you considering supporting the multi-package?

swift-kim: As you might know we already support [multi-app](https://github.com/flutter-tizen/flutter-tizen/issues/143) packages.

As you said it's perfectly fine to keep only the first element of the list because the other elements are not used at the moment, but I was not sure how to name the variable as a singular noun. (Maybe `_primaryApplication`?)

WonyoungChoi: If there is a possibility that the applications list will be used later, it may be possible to keep it. but  if you change it `_primaryApplication` is good to me.

swift-kim: It might be used in the future, e.g. to figure out if the package is a multi-app package.

WonyoungChoi (approved):
LGTM 👍

PR #310, opened 2022-01-06
Fix getArtifactPath crash when arguments are null

swift-kim:
Fixes https://github.com/flutter-tizen/flutter-tizen/issues/309.

### DIFF
--- a/lib/tizen_artifacts.dart
+++ b/lib/tizen_artifacts.dart
@@ -40,12 +40,13 @@ class TizenArtifacts extends CachedArtifacts {
     BuildMode? mode,
     EnvironmentType? environmentType,
   }) {
-    if (artifact == Artifact.genSnapshot) {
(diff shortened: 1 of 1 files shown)


HakkyuKim:
Hmm, I'm not sure what was causing the issue #309, what triggered "Null check operator used on a null value"?

swift-kim:
@HakkyuKim `mode!` causes the error here.

https://github.com/flutter-tizen/flutter-tizen/blob/ff8141bf44539d6df799cc0302c4a9b502b07a8f/lib/tizen_artifacts.dart#L52

Both `platform` and `mode` are null when `getArtifactPath` is called by https://github.com/flutter/flutter/blob/2.8.1/packages/flutter_tools/lib/src/doctor.dart#L465.

swift-kim:
I'll merge this now because the issue is urgent. I'll create another PR for further related work.

PR #311, opened 2022-01-07
Let TizenAotElf respect the target platform

swift-kim:
- Specify the `platform` argument of `Source.artifact(Artifact.genSnapshot)` so that a correct gen_snapshot artifact is resolved as a dependency of `TizenAotElf`.
- Set `environment.engineVersion` to null when a local engine is being used.
- (Refactoring) Move `getArchForTargetPlatform` and `getTargetPlatformForArch` into `tizen_build_info.dart`.

This change doesn't have any effect for normal users because source artifact paths are resolved only when using a local engine:
https://github.com/flutter/flutter/blob/2.8.1/packages/flutter_tools/lib/src/build_system/source.dart#L169-L183

Additional note (why we use both `TargetPlatform.tester` and `TargetPlatform.android*` in our code):
- `TargetPlatform.tester` is used to identify the device type (i.e. `Device.targetPlatform`).
- `TargetPlatform.android` and variants are used during the build process to represent the target architecture (arm, arm64, x86).

### DIFF
--- a/lib/build_targets/application.dart
+++ b/lib/build_targets/application.dart
@@ -94,8 +94,9 @@ abstract class TizenAssetBundle extends Target {
 ///
 /// Source: [AotElfRelease] in `common.dart`
 class TizenAotElf extends AotElfBase {
-  TizenAotElf(this.buildMode);
+  TizenAotElf(this.targetPlatform, this.buildMode);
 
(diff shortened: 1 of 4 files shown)


WonyoungChoi (approved):
LGTM

PR #312, opened 2022-01-08
Add the debug-native command

swift-kim:
Closes https://github.com/flutter-tizen/plugins/issues/295.

Changelog:
- Add a new command `debug-native` for native debugging support (especially plugin debugging). The command is experimental and hidden by default (thus not listed in [the main doc](https://github.com/flutter-tizen/flutter-tizen/blob/master/doc/commands.md)). You can find the detailed usage at [the wiki](https://github.com/flutter-tizen/flutter-tizen/wiki/Debugging-app's-native-code).
- Change the default app launch command from `app_launcher -s` to `app_launcher -e` (direct launch).
- Add the `updateLaunchJsonWithRemoteDebuggingInfo` method to `vscode_helper.dart`, which creates or updates a debug configuration in `launch.json` file.

Design considerations:
- At first I tried to add an optional switch `--debug-native` to the existing `run` command without implementing a new command, but it turned out that launching an app under gdbserver (i.e. [launch and debug](https://github.com/flutter-tizen/plugins/issues/295#issuecomment-998603229)) was too unstable and caused SIGABRT on some devices (Tizen 5.5 watch). Attaching gdbserver automatically after app launch was also not possible because it caused many bugs and inconsistent results (when is it safe to attach gdbserver to a fresh debuggee?).
- The current design requires the user to open at least two console windows for CLI debugging (one for running the `debug-native` command and the other for launching GDB). I tried to support launching GDB directly from the tool process (e.g. by pressing a key after running the `debug-native` command) and pipe their inputs and outputs (which can be easily done when I use Python), but failed after all because Dart had too limited support for child process manipulation and stream processing.

Tested on:
- Hosts: Ubuntu, Windows (PowerShell and WSL)
- Devices: Galaxy Watch Active, Wearable emulator, Mobile emulator

Please let me know if you have any suggestion or the command doesn't work as expected.

### DIFF
--- a/lib/commands/debug_native.dart
+++ b/lib/commands/debug_native.dart
@@ -49,7 +49,8 @@ class DebugNativeCommand extends FlutterCommand {
   TizenDevice _device;
   FlutterProject _project;
   TizenTpk _package;
-  StreamSubscription<void> _subscription;
+
+  final Completer<void> _finished = Completer<void>();
 
   @override
   Future<void> validateCommand() async {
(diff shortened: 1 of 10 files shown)


swift-kim:
Another option would be to create a standalone python tool just like Android's [flutter_gdb](https://github.com/flutter/engine/blob/master/sky/tools/flutter_gdb). In this case, some parts of the flutter-tizen codebase should be re-implemented in python because they can't be reused (e.g. the logic for locating the Tizen SDK).

swift-kim:
I'm currently implementing a python version of this. I'll open a new PR when it's done.

swift-kim:
Closing in favor of https://github.com/flutter-tizen/flutter-tizen/pull/315.

PR #314, opened 2022-01-11
Organize the use of globals

swift-kim:
Use globals if and only if it makes sense.

### DIFF
--- a/lib/executable.dart
+++ b/lib/executable.dart
@@ -202,8 +202,10 @@ Future<void> main(List<String> args) async {
       TizenValidator: () => TizenValidator(
             tizenSdk: tizenSdk,
             dotnetCli: dotnetCli,
+            fileSystem: globals.fs,
             logger: globals.logger,
             processManager: globals.processManager,
+            userMessages: globals.userMessages,
           ),
       TizenWorkflow: () => TizenWorkflow(
             tizenSdk: tizenSdk,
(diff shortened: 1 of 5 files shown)


lib/executable.dart
```diff
@@ -197,20 +192,15 @@ Future<void> main(List<String> args) async {
...
-            artifacts: globals.artifacts,
-            usage: globals.flutterUsage,
-            platform: globals.platform,
-          ),
+      TizenBuilder: () => TizenBuilder(),
       TizenSdk: () => TizenSdk.locateSdk(),
       TizenValidator: () => TizenValidator(
             tizenSdk: tizenSdk,
             dotnetCli: dotnetCli,
+            fileSystem: globals.fs,
```
WonyoungChoi, line 200: Can we remove these global arguments also like `TizenBuilder`? What are the criteria for removing the global arguments?

swift-kim: Good question. Explicitly feeding arguments to a class (thus not using globals) generally improves testability and clarity of the code (https://github.com/flutter/flutter/issues/47161), but that's not always the case and sometimes it's hard to say which way is better. My personal criteria is to use globals when either:

- Too many of them are used in the class so using globals makes the implementation convenient and cleaner.
- It's hard to avoid context injection in the test code or there's no plan to add tests for the class.

WonyoungChoi (approved):
LGTM

PR #317, opened 2022-01-17
Add tests for tizen_tpk.dart

swift-kim:


WonyoungChoi (approved):
LGTM, but I hope more tests will be added later to cover other exceptional cases.

swift-kim:
I tried but failed to add tests for `TizenTpk.fromTpk` because `unzip` doesn't actually occur during tests (it seemed not to be supported by `MemoryFileSystem`). For other failing cases I didn't really see the necessity to add tests because they looked too trivial.

PR #322, opened 2022-01-25
Symbolic links fix in linking process.

JRazek:
This PR fix the issue with symlinks being included into linking with their names corrupted. 
Files such as `libprotobuf-lite.so.25.0.0` were added to linking with `-lprotobuf-lite.so.25.0` - after the last dot everything was deleted and they could not be found. 
Now the symlinks are copied, but are not added into linking.

### DIFF
--- a/lib/build_targets/plugins.dart
+++ b/lib/build_targets/plugins.dart
@@ -199,9 +199,9 @@ class NativePlugins extends Target {
               if (userLibs.contains(libName)) {
                 continue;
               }
-              lib.copySync(libDir.childFile(lib.basename).path);
               userLibs.add(libName);
             }
+            lib.copySync(libDir.childFile(lib.basename).path);
 
             inputs.add(lib);
             if (isSharedLib) {
--- a/lib/build_targets/plugins.dart
+++ b/lib/build_targets/plugins.dart
@@ -188,15 +188,20 @@ class NativePlugins extends Target {
       for (final Directory directory
           in pluginLibDirs.where((Directory d) => d.existsSync())) {
         for (final File lib in directory.listSync().whereType<File>()) {
-          final bool isSharedLib = lib.basename.endsWith('.so');
+          final bool isSharedLib =
+              lib.basename.contains(RegExp(r'(\.so)((\..*)|$)'));
           final bool isStaticLib = lib.basename.endsWith('.a');
+
           if (isSharedLib || isStaticLib) {
-            final String libName = getLibNameForFileName(lib.basename);
-            if (userLibs.contains(libName)) {
-              continue;
+            final bool isLinkableSharedLib = lib.basename.endsWith('.so');
+            if (isLinkableSharedLib || isStaticLib) {
+              final String libName = getLibNameForFileName(lib.basename);
+              if (userLibs.contains(libName)) {
+                continue;
+              }
+              lib.copySync(libDir.childFile(lib.basename).path);
+              userLibs.add(libName);
             }
-            lib.copySync(libDir.childFile(lib.basename).path);
-            userLibs.add(libName);
 
             inputs.add(lib);
             if (isSharedLib) {


swift-kim:
I don't quite get what you are trying to fix. Could you elaborate on what the issue is? Why do you have symbolic links in your `lib` directory? (Note that symlinks are not portable and shouldn't be published on pub.dev.) How about just deleting the symlinks and renaming the versioned lib file to `libprotobuf-lite.so`?

JRazek:
> I don't quite get what you are trying to fix. Could you elaborate on what the issue is? Why do you have symbolic links in your `lib` directory? (Note that symlinks are not portable and shouldn't be published on pub.dev.) How about just deleting the symlinks and renaming the versioned lib file to `libprotobuf-lite.so`?

I added an issue. This PR is now related to #323

WonyoungChoi:
@swift-kim The `SONAME` of `libprotobuf-lite.so` is `libprotobuf-lite.so.25`.  The renaming to libprotobuf-lite.so can not be a solution without changing the `SONAME` by using like `patchelf`. 😢

lib/build_targets/plugins.dart
```diff
@@ -188,15 +188,20 @@ class NativePlugins extends Target {
...
-          final bool isSharedLib = lib.basename.endsWith('.so');
+          final bool isSharedLib =
+              lib.basename.contains(RegExp(r'(\.so)((\..*)|$)'));
           final bool isStaticLib = lib.basename.endsWith('.a');
+
           if (isSharedLib || isStaticLib) {
-            final String libName = getLibNameForFileName(lib.basename);
-            if (userLibs.contains(libName)) {
-              continue;
+            final bool isLinkableSharedLib = lib.basename.endsWith('.so');
```
swift-kim, line 196: Why should `libprotobuf-lite.so.25` be excluded from `userLibs`?

JRazek: In this way it is linked to `libflutter_plugins` after being edited by this function 
```dart
String getLibNameForFileName(String name) {
  if (name.startsWith('lib')) {
    name = name.substring(3);
  }
  if (name.lastIndexOf('.') > 0) {
    name = name.substring(0, name.lastIndexOf('.'));
  }
  return name;
} 
```
the ending compiler argument looks like this: `-lprotobuf-lite.so` and it should either be converted fully to lib name or linked with `-l:` - `-l:libprotobuf-lite.so.25`
```dart
 if (isLinkableSharedLib || isStaticLib) {
   final String libName = getLibNameForFileName(lib.basename);
   if (userLibs.contains(libName)) {
     continue;
   }
   userLibs.add(libName);
}
lib.copySync(libDir.childFile(lib.basename).path);
```
this code copies all the files to the device, but only uses the ones ending with `.so` to linking

swift-kim: Shouldn't the final structure of the `lib` directory be like

```
lib
├── libflutter_plugins.so
└── libprotobuf-lite.so.25
```

Do you still think the symbolic links should be included in the TPK?

JRazek: I suppose that it is not necessary to copy all the symlinks. `libprotobuf-lite.so.25` would be enough.

JRazek:
I managed to fix the issue by changing the `SONAME` of `libprotobuf.so.25.0.0` to `libprotobuf.so` with `patchelf` tool.

swift-kim:
@JRazek Yeah, that's a feasible option too. (You can set the soname at the build time without having to use the patchelf tool.) But I'll also check if we can allow other shared lib extensions than `.so`.

Please also note that you can place the lib under `lib/arm` or `lib/armel` if the lib is an arm binary and you plan to support multiple architectures (arm, arm64, and x86).

swift-kim:
I couldn't find a working solution that can work both on Unix and Windows. I tried your patch on Windows but Linux symlinks are not recognized by the linker and therefore result in a build error:

```
C:\Apps\tizen-studio\tools\smart-build-interface/../arm-linux-gnueabi-gcc-9.2/lib/gcc/arm-tizen-linux-gnueabi/9.2.0/../../../../arm-tizen-linux-gnueabi/bin\ld:D
:/Git/dummy/flutter_blue_tizen/example/.dart_tool/flutter_build/60caf926d51921ec2eaf10e3c105902d/tizen_plugins/lib/libprotobuf-lite.so: file format not
recognized; treating as linker script
C:\Apps\tizen-studio\tools\smart-build-interface/../arm-linux-gnueabi-gcc-9.2/lib/gcc/arm-tizen-linux-gnueabi/9.2.0/../../../../arm-tizen-linux-gnueabi/bin\ld:D
:/Git/dummy/flutter_blue_tizen/example/.dart_tool/flutter_build/60caf926d51921ec2eaf10e3c105902d/tizen_plugins/lib/libprotobuf-lite.so:0: syntax error
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

Thus the following limitations will remain unresolved.

- No symbolic links are allowed in plugin's `lib` directory.
- The file names and sonames of shared libs in the directory should always end with `.so`.

PR #337, opened 2022-02-28
Create only plugin registrant file that match project language

WonyoungChoi:
- Create `flutter/GeneratedPluginRegistrant.cs` if the project language is `csharp`.
- Create `flutter/generated_plugin_registrant.h` if the project language is `cpp`.

### DIFF
--- a/lib/tizen_plugins.dart
+++ b/lib/tizen_plugins.dart
@@ -265,6 +265,7 @@ Future<void> ensureReadyForTizenTooling(FlutterProject project) async {
   await tizenProject.ensureReadyForPlatformSpecificTooling();
 
   await injectTizenPlugins(project);
+  await _noticeAvailableTizenPlugins(project);
 }
 
 /// See: [injectPlugins] in `flutter_plugins.dart`
@@ -273,10 +274,11 @@ Future<void> injectTizenPlugins(FlutterProject project) async {
   if (tizenProject.existsSync()) {
     final List<TizenPlugin> nativePlugins =
         await findTizenPlugins(project, nativeOnly: true);
-    _writeCppPluginRegistrant(tizenProject.managedDirectory, nativePlugins);
-    _writeCsharpPluginRegistrant(tizenProject.managedDirectory, nativePlugins);
+    _writeTizenPluginRegistrant(tizenProject, nativePlugins);
   }
+}
 
+Future<void> _noticeAvailableTizenPlugins(FlutterProject project) async {
   final List<String> plugins =
       (await findPlugins(project)).map((Plugin p) => p.name).toList();
   for (final String plugin in plugins) {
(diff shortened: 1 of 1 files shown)


lib/tizen_plugins.dart
```diff
@@ -265,6 +265,7 @@ Future<void> ensureReadyForTizenTooling(FlutterProject project) async {
   await tizenProject.ensureReadyForPlatformSpecificTooling();
 
   await injectTizenPlugins(project);
+  await _noticeAvailableTizenPlugins(project);
```
swift-kim, line 268: "notice" means "become aware of".
```suggestion
  await _informAvailableTizenPlugins(project);
```

lib/tizen_plugins.dart
```diff
@@ -273,10 +274,11 @@ Future<void> injectTizenPlugins(FlutterProject project) async {
   if (tizenProject.existsSync()) {
     final List<TizenPlugin> nativePlugins =
         await findTizenPlugins(project, nativeOnly: true);
-    _writeCppPluginRegistrant(tizenProject.managedDirectory, nativePlugins);
-    _writeCsharpPluginRegistrant(tizenProject.managedDirectory, nativePlugins);
+    _writeTizenPluginRegistrant(tizenProject, nativePlugins);
```
swift-kim, line 277: Tizen plugin can refer to either Dart or native plugin. Let's be more specific.

```suggestion
    _writeNativePluginRegistrant(tizenProject, nativePlugins);
```

test/general/tizen_plugins_test.dart
```diff
@@ -92,7 +92,7 @@ class _PluginRegistrant {
     ProcessManager: () => FakeProcessManager.any(),
   }, testOn: 'posix');
 
-  testUsingContext('Generates native plugin registrants', () async {
+  testUsingContext('Generates native plugin registrants for cpp', () async {
```
swift-kim, line 95: Singular form should be used since only a single plugin registrant is generated.
```suggestion
  testUsingContext('Generates native plugin registrant for C++', () async {
```
Or simply
```suggestion
  testUsingContext('Generates C++ plugin registrant', () async {
```

test/general/tizen_plugins_test.dart
```diff
@@ -136,6 +136,43 @@ void RegisterPlugins(flutter::PluginRegistry *registry) {
       registry->GetRegistrarForPlugin("SomeNativePlugin"));
 }
 '''));
+  }, overrides: <Type, Generator>{
+    FileSystem: () => fileSystem,
+    ProcessManager: () => FakeProcessManager.any(),
+  }, testOn: 'posix');
+
+  testUsingContext('Generates native plugin registrants for csharp', () async {
```
swift-kim, line 144:
```suggestion
  testUsingContext('Generates native plugin registrant for C#', () async {
```

swift-kim (commented):
Otherwise looks good to me!

lib/tizen_plugins.dart
WonyoungChoi: I agree, but I wanted to sync same style with other platforms in `flutter_plugins.dart`. It will  help us refer to other platforms' implementation later. And it implies that `writingXXXXPluginRegistrant` method name is for only native plugins.

swift-kim: Ah great. That makes sense too.

WonyoungChoi: Thank you ❤️

swift-kim (approved):

PR #349, opened 2022-03-25
Remove PackageBuilder and use a unified BuildSystem

swift-kim:
- Generate the asset bundle (`flutter_assets`) in `buildDir` instead of `outputDir` to deal with the issue described in https://github.com/flutter-tizen/flutter-tizen/pull/51.
- Remove `PackageBuilder` and add `DebugTizenApplication`/`ReleaseTizenApplication` to dependencies of `DotnetTpk`/`NativeTpk`.
- Rename the target `Package` to `TizenPackage` and make it always unskippable.


HakkyuKim (approved):
:+1: for simple solution to #51, LGTM.

PR #352, opened 2022-03-30
Add MethodChannel.SetMethodCallHanlder() for sync callback

WonyoungChoi:
- Add a overloading of `MethodChannel.SetMethodCallHandler` for sync callback.
- Remove unnecessary delegate definition of `MethodCallHandler` to avoid hiding the callback method signature.
- Add `MethodChannel.UnsetMethodCallHandler()`. `ArgumentNullException` will be thrown if handler parameter of `SetMethodCallHandler()` is null.

### example:
```csharp
MethodChannel channel = new MethodChannel("foo");
channel.SetMethodCallHandler((call) => "bar");  // sync callback
channel.SetMethodCallHandler(async (call) => "bar"); // async callback
channel.SetMethodCallHandler((call) => Task.FromResult<object>("bar"));  // async callback
```

### DIFF
--- a/embedding/csharp/Tizen.Flutter.Embedding.Tests/Channels/MethodChannelTests.cs
+++ b/embedding/csharp/Tizen.Flutter.Embedding.Tests/Channels/MethodChannelTests.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Linq;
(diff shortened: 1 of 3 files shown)


swift-kim (approved):
Nice idea!

PR #358, opened 2022-04-22
Apply embedder C API changes

bbrto21:
Depends on https://github.com/flutter-tizen/engine/pull/278


bbrto21:
I'll update engine hash after https://github.com/flutter-tizen/engine/pull/278 merged

embedding/cpp/include/flutter_app.h
```diff
@@ -52,9 +52,6 @@ class FlutterApp : public flutter::PluginRegistry {
  protected:
   void ParseEngineArgs();
 
-  // Whether the app is headed or headless.
-  bool is_headed_ = true;
```
WonyoungChoi, line 56: Is there any problem with backward-compatibility if the public API is deleted without the deprecation process?

WonyoungChoi, line 56: Is there any problem with backward-compatibility if the public API is deleted without the deprecation process?

bbrto21: I thought there was no problem because I guess the only use of this is Embedding.
However, If the deprecation process is mandatory, I will. please guide me.

WonyoungChoi: This public API can be used in application runners exposed to users. If some application is using this API, there will be a problem in build. 
/cc @swift-kim

swift-kim: `is_headed_` is an internal API, not a public API. Public APIs are defined in the `public:` section (e.g. `SetDartEntrypoint`).

Also there should be no app (including our internal apps) that directly accesses the `is_headed_` value.

WonyoungChoi: The meaning of the `public api` I mentioned is all APIs exposed to the user.
In that sense, protected members also correspond to the public api. But as you said, it's very unlikely that these APIs were used.

swift-kim: Although the internal APIs are exposed to users, they are not officially supported and we don't guarantee their forward compatibility.

embedding/csharp/Tizen.Flutter.Embedding/FlutterApplication.cs
```diff
@@ -75,7 +70,8 @@ public class FlutterApplication : CoreUIApplication, IPluginRegistry
         /// <summary>
         /// The Flutter engine instance handle.
         /// </summary>
-        protected internal FlutterDesktopEngine Handle { get; private set; } = new FlutterDesktopEngine();
+        protected internal FlutterDesktopEngine FlutterEngine { get; private set; } = new FlutterDesktopEngine();
```
swift-kim, line 73: How about renaming to `Engine` for consistency with `EngineArgs` (in this file) and `engine_` (C++)?

embedding/csharp/Tizen.Flutter.Embedding/FlutterApplication.cs
```diff
@@ -75,7 +70,8 @@ public class FlutterApplication : CoreUIApplication, IPluginRegistry
...
+        protected internal FlutterDesktopView FlutterView { get; private set; } = new FlutterDesktopView();
```
swift-kim, line 74: Please add documentation.

embedding/csharp/Tizen.Flutter.Embedding/FlutterApplication.cs
```diff
@@ -122,74 +117,71 @@ protected override void OnCreate()
                     dart_entrypoint_argv = entrypointArgs.Handle,
                 };
 
-                Handle = FlutterDesktopRunEngine(ref windowProperties, ref engineProperties);
-                if (Handle.IsInvalid)
-                {
-                    throw new Exception("Could not launch a Flutter application.");
-                }
+                FlutterEngine = FlutterDesktopEngineCreate(ref engineProperties);
```
swift-kim, line 120: Shouldn't you check if the value is valid as you did in the C++ side?

embedding/csharp/Tizen.Flutter.Embedding/Interop/FlutterDesktopView.cs
```diff
@@ -0,0 +1,29 @@
...
+        public FlutterDesktopView() : base(IntPtr.Zero, true)
+        {
+        }
+
+        public override bool IsInvalid => handle == IntPtr.Zero;
+
+        public static implicit operator bool(FlutterDesktopView view)
+        {
+            return !view.IsInvalid;
+        }
```
swift-kim, line 21: This operator overloading should not be necessary at this moment.

FYI) `FlutterDesktopEngine` objects are implicitly converted to `bool` to avoid double negation in `Debug.Assert` checks.

```csharp
// Bad
Debug.Assert(!Handle.IsInvalid);
// Good
Debug.Assert(Handle);
```

embedding/csharp/Tizen.Flutter.Embedding/Interop/Interop.cs
```diff
@@ -60,28 +61,33 @@ public struct FlutterDesktopEngineProperties
...
             FlutterDesktopEngine engine);
 
         [DllImport("flutter_tizen.so")]
-        public static extern void FlutterDesktopNotifyAppIsPaused(
+        public static extern void FlutterDesktopEngineNotifyAppIsPaused(
+            FlutterDesktopEngine engine);
+
+        [DllImport("flutter_tizen.so")]
+        public static extern FlutterDesktopView FlutterDesktopViewCreateFromNewWindow(
+            ref FlutterDesktopWindowProperties window_properties,
```
swift-kim, line 90: Will you add these missing functions later when they become necessary?

- `FlutterDesktopEngineNotifyAppIsInactive`
- `FlutterDesktopEngineNotifyAppIsDetached`
- `FlutterDesktopPluginRegistrarGetNativeWindow`

embedding/cpp/include/flutter_app.h
bbrto21: Hey guys... what's the conclusion? Is it correct that the deprecation process is not required in this case?

embedding/cpp/flutter_service_app.cc
```diff
@@ -6,8 +6,41 @@
 
 #include "tizen_log.h"
 
-FlutterServiceApp::FlutterServiceApp() {
-  is_headed_ = false;
+bool FlutterServiceApp::OnCreate() {
```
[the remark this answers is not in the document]
bbrto21: All right. I also did not like the current inheritance relationship in the C++ implementation.
I will revise it according to your opinion.

embedding/cpp/include/flutter_app.h
WonyoungChoi: In this case, there seems to be no actual app to use `is_headed_`, so you can ignore the ABI Break. But for the future, it seems necessary to clarify what is internal API or not.
/cc @swift-kim

embedding/cpp/flutter_service_app.cc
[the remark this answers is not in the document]
swift-kim: Thank you.

There are still some common parts used by both classes such as `ParseEngineArgs`. How should we handle them?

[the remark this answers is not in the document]
bbrto21: Um... I gonna make a util class just like a C# embedding. how about it?

embedding/csharp/Tizen.Flutter.Embedding/Interop/Interop.cs
bbrto21: Yes, I'll add them when I need.

embedding/cpp/flutter_service_app.cc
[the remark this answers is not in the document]
swift-kim: Yes, that's possible. Another option is to create a common base class (such as `FlutterAppBase`) for the two classes. In that way, some common members (such as `engine_`) can also be inherited from the base class. It's up to your preference.

embedding/csharp/Tizen.Flutter.Embedding/Interop/FlutterDesktopView.cs
bbrto21: Thank you, I removed it.

embedding/csharp/Tizen.Flutter.Embedding/FlutterApplication.cs
bbrto21: I added check to validate handle.

embedding/cpp/include/flutter_app.h
```diff
@@ -90,7 +87,10 @@ class FlutterApp : public flutter::PluginRegistry {
   std::vector<std::string> dart_entrypoint_args_;
 
   // The Flutter engine instance handle.
-  FlutterDesktopEngineRef handle_ = nullptr;
+  FlutterDesktopEngineRef engine_ = nullptr;
+
+  // The Flutter engine instance handle.
+  FlutterDesktopViewRef view_ = nullptr;
```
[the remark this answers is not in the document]
bbrto21: I fixed it.

embedding/cpp/flutter_service_app.cc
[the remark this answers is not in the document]
bbrto21: Thanks for your reply. 
I chose to introduce Utils class. this change makes the two embeddings more similar. 
And, at least to me, this looks better.

embedding/cpp/flutter_service_app.cc
```diff
@@ -4,54 +4,124 @@
...
+
+  FlutterDesktopEngineNotifyLocaleChange(engine_);
 }
 
 int FlutterServiceApp::Run(int argc, char **argv) {
   service_app_lifecycle_callback_s lifecycle_cb = {};
   lifecycle_cb.create = [](void *data) -> bool {
-    FlutterApp *app = (FlutterApp *)data;
-    return app->OnCreate();
+    auto *service_app = reinterpret_cast<FlutterServiceApp *>(data);
```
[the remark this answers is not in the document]
bbrto21: Should I revert this change?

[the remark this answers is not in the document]
swift-kim: No, just simplifying the name will do.

embedding/cpp/include/flutter_app.h
```diff
@@ -6,6 +6,7 @@
 #define FLUTTER_TIZEN_EMBEDDING_CPP_INCLUDE_FLUTTER_APP_H_
 
 #include <app.h>
+
```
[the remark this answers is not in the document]
bbrto21: Thanks!, I found out that formatter setting of my vscode was wrong.

embedding/cpp/utils.h
```diff
@@ -0,0 +1,17 @@
...
+
+#ifndef FLUTTER_TIZEN_EMBEDDING_CPP_UTILS_H_
+#define FLUTTER_TIZEN_EMBEDDING_CPP_UTILS_H_
+
+#include <string>
+#include <vector>
+
+class Utils {
+ public:
+  // Reads engine arguments passed from the flutter-tizen tool and adds to list.
```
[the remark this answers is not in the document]
bbrto21: Where can I find the information related comment style?

Issue #365, opened 2022-05-12
Migrate to Flutter 3.0


swift-kim:
https://medium.com/flutter/introducing-flutter-3-5eb69151622f

PR #366, opened 2022-05-16
Update for Flutter 3.0

swift-kim:
Closes https://github.com/flutter-tizen/flutter-tizen/issues/365.

- Increase the framework version to 3.0.0 (https://github.com/flutter/flutter/commit/ee4e09cce01d6f2d7f4baebd247fde02e5008851).
- Update the engine hash to [3ec59ac](https://github.com/flutter-tizen/engine/releases/tag/3ec59ac). (Note: Compressed pointers are now enabled for arm64 targets.)
- Update the tool code according to the upstream side changes. Please refer to the commit history for details.


swift-kim:
I decided not to add the new "FFI plugin and template" support because it will add too much complexity without much benefit. Note that this has nothing to do with already existing FFI plugins such as url_launcher_tizen. The new FFI template consists of example C code for a shared library ([`src.tmpl`](https://github.com/flutter/flutter/tree/master/packages/flutter_tools/templates/plugin_ffi/src.tmpl)) and CMake files for building the library that we don't use for our FFI plugin implementation.

xiaowei-guan (approved):

Issue #368, opened 2022-05-17
"flutter-tizen pub get" fails on initial run


swift-kim:
How to reproduce:
```sh
$ rm -rf /path/to/flutter-tizen/flutter/bin/cache
$ flutter-tizen pub get
The current Flutter SDK version is 0.0.0-unknown.

Because wakelock_tizen requires Flutter SDK version >=2.0.0, version solving failed.
```

This results in CI failures in the plugins repo. ([example](https://github.com/flutter-tizen/plugins/runs/6469736931?check_suite_focus=true))

Related issue: https://github.com/flutter/flutter/issues/18532

Possible workarounds include:

- a) Run `flutter-tizen --version` before running `flutter-tizen pub get`.
- b) Deep clone the Flutter SDK by removing `--depth=1` in: https://github.com/flutter-tizen/flutter-tizen/blob/7dac52615868ccdf5ea46efdffc0e927dd7bf6b5/bin/internal/shared.sh#L40

Cloning time comparison:

```sh
$ time git clone https://github.com/flutter/flutter.git
...
real    0m23.405s
user    0m11.328s
sys     0m5.117s

$ time git clone --depth=1 https://github.com/flutter/flutter.git
...
real    0m4.712s
user    0m0.570s
sys     0m0.436s
```

Note: The time depends on the network condition and etc.

swift-kim:
This looks like a side effect of [this commit](https://github.com/flutter-tizen/flutter-tizen/pull/366/commits/1f2a012dde10976c9db2584699d555d3f78760fd) in the recent PR https://github.com/flutter-tizen/flutter-tizen/pull/366.

The `shared.sh` script used to run `flutter --version` every time `flutter-tizen` was run (the first solution in the above) but it doesn't do that any more.

PR #369, opened 2022-05-17
Do not shallow clone the Flutter repo

swift-kim:
Fixes https://github.com/flutter-tizen/flutter-tizen/issues/368.

Effectively there should be no regression in the total fetch time because `flutter --version` also fetches all tags by default. Actually `git clone --depth=1` + `git fetch --tags` takes longer than a single `git clone`.


HakkyuKim (approved):

wanchao-xu (approved):
Looks good to me

WonyoungChoi (approved):
LGTM :)

PR #378, opened 2022-06-20
Add Pause/Resume to ElmFlutterView

bbrto21:
Signed-off-by: Boram Bae <boram21.bae@samsung.com>

### DIFF
--- a/embedding/cpp/elm_flutter_view.cc
+++ b/embedding/cpp/elm_flutter_view.cc
@@ -4,6 +4,8 @@
 
 #include "include/elm_flutter_view.h"
 
+#include <cassert>
+
 #include "tizen_log.h"
 #include "utils.h"
 
@@ -59,6 +61,16 @@ void ElmFlutterView::Resize(int32_t width, int32_t height) {
     height_ = height;
     FlutterDesktopViewResize(view_, width_, height_);
   }
+}
+
+void ElmFlutterView::Pause() {
+  assert(IsRunning());
+  FlutterDesktopEngineNotifyAppIsPaused(engine_);
+}
+
+void ElmFlutterView::Resume() {
+  assert(IsRunning());
+  FlutterDesktopEngineNotifyAppIsResumed(engine_);
 }
 
 FlutterDesktopPluginRegistrarRef ElmFlutterView::GetRegistrarForPlugin(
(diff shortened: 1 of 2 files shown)


swift-kim:
The user should also be able to "shutdown" the engine or "detach" it from the host view.

bbrto21:
That's right, I'm already thinking about attach and detach. but I haven't been able to define the behavior of these myself yet.

bbrto21:
> The user should also be able to "shutdown" the engine or "detach" it from the host view.

By the way, I try to make PR as small as possible :)

JSUYA:
What about dependent on `SHOW` and `HIDE` events of `evas_object_`?
I think this can follow the policy of the application or uifw without a separate Pause and Resume setting.
(Of course I can't be sure that all evas_objects are `HIDE`d when the App is Paused)

bbrto21:
> What about dependent on `SHOW` and `HIDE` events of `evas_object_`?

That's a good idea. If I understand you correctly, it seems that we have two options. 
And you seem to be talking about option 2. right?

1. Provide manual APIs (current PR).
2. Provide the view behaviors dependent on Event.

JSUYA:
> > What about dependent on `SHOW` and `HIDE` events of `evas_object_`?
> 
> That's a good idea. If I understand you correctly, it seems that we have two options. And you seem to be talking about option 2. right?
> 
> 1. Provide manual APIs (current PR).
> 2. Provide behaviors of the view dependent on Event.

my opinion is 2. Um..
There may be a way to use window's EFL_APP_EVENT_RESUME and EFL_APP_EVENT_PAUSE.
(I did not check whether these events work well)

bbrto21:
> The user should also be able to "shutdown" the engine or "detach" it from the host view.

@swift-kim
`shutdown engine` follows when the view is destroyed now. Is this enough?
I'm still investigating how to `detach` the `view` from the `host`(or `parent`). do you know anything about this?

swift-kim:
> Is this enough?

Yes, I think we're handling it correctly. (Or do we need an explicit `Destroy` method?)

> do you know anything about this?

I don't have much idea but if there's no API for re-parenting a Elementary widget (table), there can only be a workaround for the problem at best.

bbrto21:
the engine getter was added to ElmFlutterView(#421), therefore, it seems better to control the life cycle through the engine. so I close this draft.

PR #381, opened 2022-06-21
ElmFlutterView: Destory view instance

JSUYA:
When ElmFlutterView is deleted, view_ is no longer used, so destory it.

### DIFF
--- a/embedding/cpp/elm_flutter_view.cc
+++ b/embedding/cpp/elm_flutter_view.cc
@@ -9,6 +9,12 @@
 #include "include/flutter_engine.h"
 #include "tizen_log.h"
 
+ElmFlutterView::~ElmFlutterView() {
+  if (view_) {
+    FlutterDesktopViewDestroy(view_);
+  }
+}
+
 bool ElmFlutterView::RunEngine() {
   if (!parent_) {
     TizenLog::Error("The parent object is invalid.");
@@ -38,9 +44,10 @@ bool ElmFlutterView::RunEngine() {
   evas_object_ =
       static_cast<Evas_Object *>(FlutterDesktopViewGetEvasObject(view_));
   if (!evas_object_) {
-    TizenLog::Error("Could not get an Evas object");
+    TizenLog::Error("Could not get an Evas object.");
     return false;
   }
+
   return true;
 }
 
--- a/embedding/cpp/include/elm_flutter_view.h
+++ b/embedding/cpp/include/elm_flutter_view.h
@@ -24,7 +24,7 @@ class ElmFlutterView : public flutter::PluginRegistry {
       : parent_(parent),
         initial_width_(initial_width),
         initial_height_(initial_height) {}
-  virtual ~ElmFlutterView() {}
+  virtual ~ElmFlutterView();
 
   FlutterDesktopPluginRegistrarRef GetRegistrarForPlugin(
       const std::string &plugin_name) override;


embedding/cpp/elm_flutter_view.cc
```diff
@@ -16,8 +16,8 @@ bool ElmFlutterView::RunEngine() {
   // Read engine arguments passed from the tool.
   Utils::ParseEngineArgs(&engine_args_);
 
-  std::vector<const char *> switches;
-  for (auto &arg : engine_args_) {
+  std::vector<const char*> switches;
+  for (auto& arg : engine_args_) {
```
swift-kim, line 20: Do you know what causes these style changes? Obviously the config file (`.clang-format`) hasn't changed.

embedding/cpp/elm_flutter_view.cc
```diff
@@ -45,11 +45,23 @@ bool ElmFlutterView::RunEngine() {
...
   evas_object_ =
-      static_cast<Evas_Object *>(FlutterDesktopViewGetEvasObject(view_));
+      static_cast<Evas_Object*>(FlutterDesktopViewGetEvasObject(view_));
   if (!evas_object_) {
     TizenLog::Error("Could not get an Evas object");
     return false;
   }
+
+  evas_object_event_callback_add(
+      evas_object_, EVAS_CALLBACK_DEL,
```
swift-kim, line 55: When does the object get deleted? By who?

embedding/cpp/elm_flutter_view.cc
```diff
@@ -45,11 +45,23 @@ bool ElmFlutterView::RunEngine() {
   }
 
   evas_object_ =
-      static_cast<Evas_Object *>(FlutterDesktopViewGetEvasObject(view_));
+      static_cast<Evas_Object*>(FlutterDesktopViewGetEvasObject(view_));
   if (!evas_object_) {
     TizenLog::Error("Could not get an Evas object");
```
swift-kim, line 50: Always.
```suggestion
    TizenLog::Error("Could not get an Evas object.");
```

embedding/cpp/elm_flutter_view.cc
JSUYA: evas_object_ is deleted when parent_ is deleted.
This deletion is process by Eo of EFL. (ElmFlutterView always needs a valid evas object (parent).)
Of course, if the user deletes FlutterView, I think that another engine problem(crash?) may occur.
I am considering several cases.

embedding/cpp/elm_flutter_view.cc
JSUYA: done

embedding/cpp/elm_flutter_view.cc
JSUYA: oh... 
All code has been changed according to the code format in the lambda function.
If I write the code below and format it, all symbols will be left-aligned.
```cpp
  evas_object_event_callback_add(
      evas_object_, EVAS_CALLBACK_DEL,
      [](void* data, Evas* evas, Evas_Object* object, void* event_info) {
        auto* self = reinterpret_cast<ElmFlutterView *>(data);
        if (self->engine_) {
          FlutterDesktopEngineShutdown(self->engine_);
          self->engine_ = nullptr;
        }
      },
      this);
```
However, if I write `auto *self ` , only the format of the argument is changed as we expected.

bbrto21:
Q. If the container created inside the view is deleted by some how on outside of view, and we know that, should the state of the view be considered valid?

embedding/cpp/include/elm_flutter_view.h
```diff
@@ -9,20 +9,25 @@
...
 class ElmFlutterView : public flutter::PluginRegistry {
  public:
-  explicit ElmFlutterView(Evas_Object *parent) : parent_(parent) {}
-  explicit ElmFlutterView(Evas_Object *parent, int32_t initial_width,
-                          int32_t initial_height)
-      : parent_(parent),
-        initial_width_(initial_width),
-        initial_height_(initial_height) {}
+  static std::unique_ptr<ElmFlutterView> Create(Evas_Object *parent);
+  static std::unique_ptr<ElmFlutterView> Create(Evas_Object *parent,
```
bbrto21, line 22: Add a blank line between two Create methods :)

JSUYA: thx :)

bbrto21:
I opened a PR to add an API that destroys the view. @JSUYA  why don't you apply that in this PR?

bbrto21 (approved):

embedding/cpp/elm_flutter_view.cc
```diff
@@ -49,9 +49,20 @@ bool ElmFlutterView::RunEngine() {
...
+    TizenLog::Error("Could not get an Evas object.");
     return false;
   }
+
+  evas_object_event_callback_add(
+      evas_object_, EVAS_CALLBACK_DEL,
+      [](void *data, Evas *evas, Evas_Object *object, void *event_info) {
+        auto engine = reinterpret_cast<FlutterDesktopEngineRef>(data);
+        if (engine) {
+          FlutterDesktopEngineShutdown(engine);
```
swift-kim, line 61: Might be updated based on https://github.com/flutter-tizen/engine/pull/310.

JSUYA: I changed it to add view_ destroy call to the destructor.

swift-kim (approved):

swift-kim:
@bbrto21 @JSUYA This PR is causing a native app build failure. Please either revert this PR or update the engine hash right away.

bbrto21:
@swift-kim I'm so sorry, I added a commit for updating engine hash to https://github.com/flutter-tizen/flutter-tizen/pull/398

PR #382, opened 2022-06-22
Introduce FlutterEngine

bbrto21:
* The FlutterEngine holds the FlutterDesktopEngineRef and wraps the associated C-API.
* FlutterEngine may relinquish the ownership of FlutterDesktopEngineRef as needed.
  it's typical example is creating a FlutterDesktopViewRef. 
* Because FlutterDesktopViewRef takes the ownership of FlutterDesktopViewRef internally when it is created.

example)
```c++
    flutter_view = new ElmFlutterView(box, 720, 400);
    auto engine = FlutterEngine::Create();
    flutter_view->SetEngine(std::move(engine));
    if (flutter_view->RunEngine()) {
      RegisterPlugins(flutter_view2);
      Evas_Object *evas_image2 = (Evas_Object *)flutter_view->evas_object();
      elm_box_pack_end(box, evas_image2);
    }
```


embedding/cpp/flutter_app.cc
```diff
@@ -11,30 +11,9 @@
...
-  engine_prop.switches = switches.data();
-  engine_prop.switches_count = switches.size();
-  engine_prop.entrypoint = dart_entrypoint_.c_str();
-  engine_prop.dart_entrypoint_argc = entrypoint_args.size();
-  engine_prop.dart_entrypoint_argv = entrypoint_args.data();
-
-  engine_ = FlutterDesktopEngineCreate(engine_prop);
+  engine_ = FlutterEngine::Create("../res/flutter_assets", "../res/icudtl.dat",
+                                  "../lib/libapp.so", dart_entrypoint_,
+                                  dart_entrypoint_args_);
```
JSUYA, line 16: dart_entrypoint_ and dart_entrypoint_args_ in FlutterApp class are currently not set anywhere
Since getting Args from ParseEngineArgs is already implemented in FlutterEngine, it is better to delete the member variable and this argument before adding a setter getter for a separate entrypoint.

embedding/cpp/flutter_engine.cc
```diff
@@ -0,0 +1,119 @@
...
+  std::vector<const char*> entrypoint_args;
+  if (dart_entrypoint_args.has_value()) {
+    for (const std::string& arg : dart_entrypoint_args.value()) {
+      entrypoint_args.push_back(arg.c_str());
+    }
+  }
+  engine_prop.dart_entrypoint_argc = entrypoint_args.size();
+  engine_prop.dart_entrypoint_argv = entrypoint_args.data();
+
+  engine_ = FlutterDesktopEngineCreate(engine_prop);
```
JSUYA, line 50: Previously, when FlutterDesktopEngineCreate() failed, an error log was output. I think we need an exception handling for this.

embedding/cpp/flutter_app.cc
bbrto21: Thank you for review, here is my answer.

> dart_entrypoint_ and dart_entrypoint_args_ in FlutterApp class are currently not set anywhere

`FlutterApp` and `FlutterService` provide them as protected members. The host implementations(including tempate) that inherit from these base classes(`FlutterApp` or `FlutterService`) can use them directly such as window offset.

> Since getting Args from ParseEngineArgs is already implemented in FlutterEngine,

The engine arguments is delivered by the tool(flutter-tizen), but not the dart arguments. so it's better to be presented as a programmable interface as it is now.

> this argument before adding a setter getter for a separate entrypoint.

I think It is better to deliver them when you create them because they correspond to the current CAPI.

embedding/cpp/flutter_engine.cc
bbrto21: You're right. I'll add it.

embedding/cpp/include/flutter_engine.h
```diff
@@ -0,0 +1,109 @@
...
+ public:
+  virtual ~FlutterEngine();
+
+  static std::unique_ptr<FlutterEngine> Create(
+      const std::string& assets_path, const std::string& icu_data_path,
+      const std::string& aot_library_path,
+      const std::optional<std::string>& dart_entrypoint = std::nullopt,
+      const std::optional<std::vector<std::string>>& dart_entrypoint_args =
+          std::nullopt);
+
```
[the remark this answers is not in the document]
bbrto21: OK. I'll add it.

bbrto21:
@JSUYA Please review again.

embedding/cpp/include/flutter_engine.h
```diff
@@ -0,0 +1,112 @@
...
+#include <vector>
+
+// The engine for flutter execution.
+class FlutterEngine : public flutter::PluginRegistry {
+ public:
+  virtual ~FlutterEngine();
+  static std::unique_ptr<FlutterEngine> Create(
+      const std::optional<std::string>& dart_entrypoint = std::nullopt,
+      const std::optional<std::vector<std::string>>& dart_entrypoint_args =
+          std::nullopt);
```
swift-kim, line 24: The use of `std::optional` seems redundant since an empty `dart_entrypoint` already indicates that the value will default to `main` ([doc](https://github.com/flutter-tizen/flutter-tizen/blob/47918330dd65e407c988f91349f16a71ef5217cb/embedding/cpp/include/flutter_app.h#L80-L81)). One option is to just revert b89fdd08e116430d41260369abbe9eb282b70a54. The other options will involve significant changes to the embedder API so we might consider them later. One of such options is to pass `FlutterDesktopEngineProperties` on `EngineRun` (instead of `EngineCreate`), but in that case we need a way to properly deal with logging options (`--verbose-logging` and `--tizen-logging-port`).

embedding/cpp/flutter_engine.cc
```diff
@@ -0,0 +1,134 @@
...
+    const std::string& aot_library_path,
+    const std::optional<std::string>& dart_entrypoint,
+    const std::optional<std::vector<std::string>>& dart_entrypoint_args) {
+  FlutterDesktopEngineProperties engine_prop = {};
+  engine_prop.assets_path = assets_path.c_str();
+  engine_prop.icu_data_path = icu_data_path.c_str();
+  engine_prop.aot_library_path = aot_library_path.c_str();
+
+  // Read engine arguments passed from the tool.
+  Utils::ParseEngineArgs(&engine_arguments_);
```
swift-kim, line 45: This function is now only used here so it might be moved from `utils.h` to this class.

embedding/cpp/flutter_engine.cc
```diff
@@ -0,0 +1,134 @@
...
+  std::vector<const char*> switches;
+  for (auto& arg : engine_arguments_) {
```
swift-kim, line 47: nit:
```suggestion
  for (const std::string& arg : engine_arguments_) {
```

embedding/cpp/flutter_engine.cc
```diff
@@ -0,0 +1,134 @@
...
+bool FlutterEngine::Run() {
+  if (engine_) {
+    is_running_ = FlutterDesktopEngineRun(engine_);
+  }
+  return is_running_;
+}
+
+void FlutterEngine::Shutdown() {
+  if (engine_) {
+    FlutterDesktopEngineShutdown(engine_);
```
swift-kim, line 83: Maybe we should set `is_running_` to false (or even set `engine_` to nullptr) after terminating the engine (although it might not be very necessary).

embedding/cpp/flutter_service_app.cc
```diff
@@ -12,34 +12,15 @@
...
-  engine_prop.switches = switches.data();
-  engine_prop.switches_count = switches.size();
-  engine_prop.entrypoint = dart_entrypoint_.c_str();
-  engine_prop.dart_entrypoint_argc = entrypoint_args.size();
-  engine_prop.dart_entrypoint_argv = entrypoint_args.data();
-
-  engine_ = FlutterDesktopEngineCreate(engine_prop);
+  engine_ = FlutterEngine::Create("../res/flutter_assets", "../res/icudtl.dat",
+                                  "../lib/libapp.so", dart_entrypoint_,
+                                  dart_entrypoint_args_);
```
swift-kim, line 17: I think you forgot to update this part.

embedding/cpp/include/elm_flutter_view.h
```diff
@@ -38,12 +41,14 @@ class ElmFlutterView : public flutter::PluginRegistry {
 
   int32_t GetHeight();
 
+  void SetEngine(std::unique_ptr<FlutterEngine> engine);
+
  private:
   // The switches to pass to the Flutter engine.
   std::vector<std::string> engine_args_;
```
swift-kim, line 48: Looks like a leftover.

embedding/cpp/include/flutter_engine.h
```diff
@@ -0,0 +1,112 @@
...
+
+#include <memory>
+#include <optional>
+#include <string>
+#include <vector>
+
+// The engine for flutter execution.
+class FlutterEngine : public flutter::PluginRegistry {
+ public:
+  virtual ~FlutterEngine();
```
swift-kim, line 20: Please add an empty line between this line and `Create`.

embedding/cpp/include/flutter_engine.h
```diff
@@ -0,0 +1,112 @@
...
+  // the Flutter app lifecycle.
+  void NotifyAppIsDetached();
+
+  // Notifies that the host app received an app control.
+  //
+  // This method sends the app control to Flutter over the "app control event
+  // channel".
+  void NotifyAppControl(app_control_h app_control);
+
+  // Notifies that low memory warning.
```
swift-kim, line 75:
```suggestion
  // Notifies that a low memory warning has been received.
```

embedding/cpp/include/flutter_engine.h
```diff
@@ -0,0 +1,112 @@
...
+#include <memory>
+#include <optional>
+#include <string>
+#include <vector>
+
+// The engine for flutter execution.
+class FlutterEngine : public flutter::PluginRegistry {
+ public:
+  virtual ~FlutterEngine();
+  static std::unique_ptr<FlutterEngine> Create(
```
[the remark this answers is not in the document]
bbrto21: Private constructor and providing a creation method that returns pointer make that object can only be dynamically created.
also, if the return type is `unique_ptr`, it strongly suggests to the user that only one pointer must exist for instance.
so the method that takes this as an argument also uses `unique_ptr`. and as you can see from the code, it can return `nullptr` to the user if a just created instance is not valid.

embedding/cpp/elm_flutter_view.cc
```diff
@@ -6,6 +6,7 @@
 
 #include <cassert>
 
+#include "include/flutter_engine.h"
 #include "tizen_log.h"
 #include "utils.h"
```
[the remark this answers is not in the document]
bbrto21: Thank you. I removed it at all place.

embedding/cpp/flutter_engine.cc
bbrto21: I moved `ParseEngineArgs` to this class.

embedding/cpp/flutter_engine.cc
bbrto21: done.

embedding/cpp/flutter_engine.cc
bbrto21: done.

embedding/cpp/include/elm_flutter_view.h
bbrto21: Oops!, I removed it.

embedding/cpp/include/flutter_engine.h
bbrto21: Thank you for your suggestion. I applied it.

embedding/cpp/include/flutter_engine.h
bbrto21: Done.

embedding/cpp/flutter_service_app.cc
bbrto21: Done.

embedding/cpp/flutter_engine.cc
swift-kim:
```cpp
std::vector<std::string> engine_args = ParseEngineArgs();

std::vector<const char*> switches;
for (...) {
```
or even
```cpp
for (const std::string& arg : ParseEngineArgs()) {
  ...
```
seems more appropriate in the current structure.

embedding/cpp/flutter_engine.cc
swift-kim: Thanks. Setting `engine_` to nullptr also affects other methods in this class so please be careful.

embedding/cpp/flutter_engine.cc
bbrto21: I chose the first proposal.

embedding/cpp/include/flutter_engine.h
bbrto21: I replaced it with the default argument.

embedding/cpp/flutter_engine.cc
bbrto21: Thank you. I've added more guards.

embedding/cpp/flutter_engine.cc
```diff
@@ -12,19 +12,20 @@
 
 namespace {
 
-void ParseEngineArgs(std::vector<std::string>* list) {
+std::vector<std::string> ParseEngineArgs() {
+  std::vector<std::string> engine_arguments;
```
swift-kim, line 16: For local consistency:
```suggestion
  std::vector<std::string> engine_args;
```
It can also be named `result` or something so that the reader can better understand the meaning.

embedding/cpp/flutter_engine.cc
```diff
@@ -72,32 +73,30 @@ std::unique_ptr<FlutterEngine> FlutterEngine::Create(
...
+    const std::string& aot_library_path, const std::string& dart_entrypoint,
+    const std::vector<std::string>& dart_entrypoint_args) {
   FlutterDesktopEngineProperties engine_prop = {};
   engine_prop.assets_path = assets_path.c_str();
   engine_prop.icu_data_path = icu_data_path.c_str();
   engine_prop.aot_library_path = aot_library_path.c_str();
 
   // Read engine arguments passed from the tool.
-  ParseEngineArgs(&engine_arguments_);
+  std::vector<std::string> engine_arguments = ParseEngineArgs();
```
swift-kim, line 84: ditto

embedding/cpp/include/flutter_engine.h
```diff
@@ -0,0 +1,108 @@
...
+
+// The engine for flutter execution.
+class FlutterEngine : public flutter::PluginRegistry {
+ public:
+  virtual ~FlutterEngine();
+
+  static std::unique_ptr<FlutterEngine> Create(
+      const std::string& dart_entrypoint = "",
+      const std::vector<std::string>& dart_entrypoint_args = {});
+  static std::unique_ptr<FlutterEngine> Create(
```
swift-kim, line 25: nit: Adding a blank line between the two `Create` methods might visually help readability.

swift-kim (approved):
Many thanks!

JSUYA (approved):
Looks good to me

PR #379, opened 2022-06-20
Add the ElmFlutterView's width/height getter methods

JSUYA:
Add an API that returns the width and height size of evas_object.
and when resizing, compare the actual size, not the initial size.

PR #384, opened 2022-06-29
Add dependencies for native host

bbrto21:
Fixes the build failure mentioned in https://github.com/flutter-tizen/flutter-tizen/pull/379#issuecomment-1169421471.

### DIFF
--- a/lib/build_targets/package.dart
+++ b/lib/build_targets/package.dart
@@ -295,6 +295,8 @@ class NativeTpk extends TizenPackage {
       'capi-appfw-app-common',
       'capi-appfw-application',
       'dlog',
+      'elementary',
+      'evas',
     ];
 
     final Directory buildDir = projectDir.childDirectory(buildConfig);


lib/build_targets/package.dart
```diff
@@ -295,6 +295,8 @@ class NativeTpk extends TizenPackage {
       'capi-appfw-app-common',
       'capi-appfw-application',
       'dlog',
+      'elementary',
```
swift-kim, line 298: Is `elementary` for future use?

swift-kim (approved):

lib/build_targets/package.dart
bbrto21: Since Elemental.h is explicitly included in elm_flutter_view.h, I added it. should I except it?

swift-kim: Not really. It might be useful in the future.

bbrto21:
We can't build a flutter-tizen app using native app host now, so I want to merge it right away. can I do that?

PR #385, opened 2022-07-01
build_targets: Add linker flag for link order dependency

JSUYA:
There is an order dependency between libembedding_cpp and flutter_tizen_{profile}.
Adding this flag will be searched repeatedly until all possible references are resolved.

refer to : https://linux.die.net/man/1/ld

### DIFF
--- a/lib/build_targets/package.dart
+++ b/lib/build_targets/package.dart
@@ -332,6 +332,7 @@ class NativeTpk extends TizenPackage {
       // The extra quotation marks ("") for linker flags are required due to
       // https://github.com/flutter-tizen/flutter-tizen/issues/218.
       '"-Wl,--unresolved-symbols=ignore-in-shared-libs"',
+      '"-Wl,--start-group"',
       '-lflutter_tizen_${buildInfo.deviceProfile}',
       '-L${libDir.path.toPosixPath()}',
       '-I${clientWrapperDir.childDirectory('include').path.toPosixPath()}',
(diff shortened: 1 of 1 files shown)


JSUYA:
[Test]
1. create tizen cpp app
2. modify runner.cc
```
#include "runner.h"

#include "generated_plugin_registrant.h"

class App {
 public:
  bool OnCreate() { return false; }
};

int main(int argc, char *argv[]) {
  App app;
  return app.OnCreate();
}

```

JSUYA:
This flag is costly. so we can also change the order between the two libraries.

JSUYA:
Or There is a way to remove `--whole-archive`. Does anyone know why this is necessary?

swift-kim:
@JSUYA I don't exactly remember but probably I added it because linking failed without it. You can remove it if it's not the case.

JSUYA:
> @JSUYA I don't exactly remember but probably I added it because linking failed without it. You can remove it if it's not the case.

I made new pr https://github.com/flutter-tizen/flutter-tizen/pull/385 please check this :)

JSUYA:
@swift-kim sorry my tests(remove --whole-archive) are wrong. It was causing problems in some cases. (ElmFlutterView sample). so #386 didn't help.

JSUYA:
> @swift-kim sorry my tests(remove --whole-archive) are wrong. It was causing problems in some cases. (ElmFlutterView sample). so #386 didn't help.

ignore this comment

PR #386, opened 2022-07-01
build_targets: Remove --whole-archive flag and realign libs

JSUYA:
The target library that needs to be referenced must be located behind it.
So change the order, --whole-archive is not needed anymore.


JSUYA:
https://github.com/flutter-tizen/flutter-tizen/pull/385

bbrto21 (approved):

swift-kim (approved):

PR #387, opened 2022-07-04
Pass command line arguments through to Dart code

swift-kim:
Add plumbing for command line arguments so that they are available inside Dart's `main`.

```dart
void main(List<String> args) {
  print(args);

  runApp(...);
}
```

Result:
```
flutter: [`zaybxcwdveuftgsh`, __AUL_TARGET_UID__, KAAAAAEEAAATAAAAX19BVUxfVEFSR0VUX1VJRF9fAAUAAAA1MDAxAA==, __AUL_STARTTIME__, MwAAAAEEAAASAAAAX19BVUxfU1RBUlRUSU1FX18AEQAAADM2OTY5MC81NDQzMjk3MDAA, __APP_SVC_PKG_NAME__, PAAAAAEEAAAVAAAAX19BUFBfU1ZDX1BLR19OQU1FX18AFwAAAGNvbS5leGFtcGxlLm5hdGl2ZV9hcHAA, __AUL_EX_CHECK_IS_WAS_APP, LQAAAAEEAAAaAAAAX19BVUxfRVhfQ0hFQ0tfSVNfV0FTX0FQUAADAAAATm8A, __AUL_CALLER_PID__, KQAAAAEEAAATAAAAX19BVUxfQ0FMTEVSX1BJRF9fAAYAAAAxODMzMgA=, __AUL_CALLER_UID__, KAAAAAEEAAATAAAAX19BVUxfQ0FMTEVSX1VJRF9fAAUAAAA1MDAxAA==, __AUL_HWACC__, IgAAAAEEAAAOAAAAX19BVUxfSFdBQ0NfXwAEAAAAVVNFAA==, __AUL_ROOT_PATH__, RwAAAAEEAAASAAAAX19BVUxfUk9PVF9QQVRIX18AJQAAAC9vcHQvdXNyL2FwcHMvY29tLmV4YW1wbGUubmF0aXZlX2FwcAA=, __AUL_INTERNAL_POOL__, LAAAAAEEAAAWAAAAX19BVUxfSU5URVJOQUxfUE9PTF9fAAYAAABmYWxzZQA=, __AUL_API_VERSION__, KAAAAAEEAAAUAAAAX19BVUxfQVBJX1ZFUlNJT05fXwAEAAAANC4wAA==, __AUL_IS_GLOBAL__, JwAAAAEEAAASAAAAX19BVUxfSVNfR0xPQkFMX18ABQAAAHRydWUA, __AUL_INSTALLED_STORAGE__, MwAAAAEEAAAaAAAAX19BVUxfSU5TVEFMTEVEX1NUT1JBR0VfXwAJAAAAaW50ZXJuYWwA, __AUL_LOADER_ID__, JAAAAAEEAAASAAAAX19BVUxfTE9BREVSX0lEX18AAgAAADEA]
```

### DIFF
--- a/embedding/cpp/flutter_app.cc
+++ b/embedding/cpp/flutter_app.cc
@@ -72,7 +72,8 @@ void FlutterApp::OnRegionFormatChanged(app_event_info_h event_info) {
 }
 
 int FlutterApp::Run(int argc, char **argv) {
-  for (int i = 0; i < argc; i++) {
+  // Skip the first argument as it's the binary name.
+  for (int i = 1; i < argc; i++) {
     dart_entrypoint_args_.push_back(argv[i]);
   }
 
--- a/embedding/cpp/flutter_app.cc
+++ b/embedding/cpp/flutter_app.cc
@@ -72,6 +72,10 @@ void FlutterApp::OnRegionFormatChanged(app_event_info_h event_info) {
 }
 
 int FlutterApp::Run(int argc, char **argv) {
+  for (int i = 0; i < argc; i++) {
+    dart_entrypoint_args_.push_back(argv[i]);
+  }
+
   ui_app_lifecycle_callback_s lifecycle_cb = {};
   lifecycle_cb.create = [](void *data) -> bool {
     auto *app = reinterpret_cast<FlutterApp *>(data);
--- a/embedding/cpp/flutter_engine.cc
+++ b/embedding/cpp/flutter_engine.cc
@@ -96,7 +96,6 @@ FlutterEngine::FlutterEngine(
   for (const std::string& arg : dart_entrypoint_args) {
     entrypoint_args.push_back(arg.c_str());
   }
-
   engine_prop.dart_entrypoint_argc = entrypoint_args.size();
   engine_prop.dart_entrypoint_argv = entrypoint_args.data();
 
--- a/embedding/cpp/flutter_service_app.cc
+++ b/embedding/cpp/flutter_service_app.cc
@@ -51,7 +51,8 @@ void FlutterServiceApp::OnRegionFormatChanged(app_event_info_h event_info) {
 }
 
 int FlutterServiceApp::Run(int argc, char **argv) {
-  for (int i = 0; i < argc; i++) {
+  // Skip the first argument as it's the binary name.
+  for (int i = 1; i < argc; i++) {
     dart_entrypoint_args_.push_back(argv[i]);
   }
 
--- a/embedding/cpp/flutter_service_app.cc
+++ b/embedding/cpp/flutter_service_app.cc
@@ -51,6 +51,10 @@ void FlutterServiceApp::OnRegionFormatChanged(app_event_info_h event_info) {
 }
 
 int FlutterServiceApp::Run(int argc, char **argv) {
+  for (int i = 0; i < argc; i++) {
+    dart_entrypoint_args_.push_back(argv[i]);
+  }
+
   service_app_lifecycle_callback_s lifecycle_cb = {};
   lifecycle_cb.create = [](void *data) -> bool {
     auto *app = reinterpret_cast<FlutterServiceApp *>(data);
--- a/embedding/csharp/Tizen.Flutter.Embedding/FlutterApplication.cs
+++ b/embedding/csharp/Tizen.Flutter.Embedding/FlutterApplication.cs
@@ -79,6 +79,8 @@ namespace Tizen.Flutter.Embedding
 
         public override void Run(string[] args)
         {
+            DartEntrypointArgs.AddRange(args);
+
             // Log any unhandled exception.
             AppDomain.CurrentDomain.UnhandledException += (s, e) =>
             {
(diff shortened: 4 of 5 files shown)


WonyoungChoi:
is there a use case that needs to process the arguments in the dart?

bbrto21:
> Result:
> 
> ```
> flutter: [/opt/usr/apps/com.example.native_app/bin/runner, `zaybxcwdveuftgsh`, __AUL_TARGET_UID__, KAAAAAEEAAAT
> ```

Is this just a printing example? or are you saying that main's args of the host application should be used as dart args?

embedding/cpp/flutter_app.cc
```diff
@@ -72,6 +72,10 @@ void FlutterApp::OnRegionFormatChanged(app_event_info_h event_info) {
 }
 
 int FlutterApp::Run(int argc, char **argv) {
+  for (int i = 0; i < argc; i++) {
+    dart_entrypoint_args_.push_back(argv[i]);
```
bbrto21, line 76: Why don't you make a setter like a `SetDartEntrypoint`?

swift-kim:
@WonyoungChoi Currently there's no specific use case, but I just thought that some developers might find this feature useful.

@bbrto21 Hmm? It's just a console output for the above sample code. I can't quite get what you mean.

JSUYA (commented):
I think `dart_entrypoint_args_` should store the arguments passed through `--dart-entrypoint-args`. Shouldn't main's args be stored in a separate variable?

embedding/cpp/flutter_app.cc
swift-kim: `dart_entrypoint_args_` (which will be changed to a private member in https://github.com/flutter-tizen/flutter-tizen/pull/388) is implicitly set by the embedding and users are not expected to access the value.

swift-kim:
@JSUYA Actually I haven't noticed there's the `--dart-entrypoint-args` option in `flutter-tizen run`, but unfortunately the option cannot be supported on Tizen because there's no way to pass the arguments (`debuggingOptions.dartEntrypointArgs`) to the app in the [`TizenDevice.startApp`](https://github.com/flutter-tizen/flutter-tizen/blob/940c43a8a2577f6efe395f565e061ffd21120d10/lib/tizen_device.dart#L322-L469) function. The `app_launcher -e` command (or the corresponding secure command) doesn't support passing extra arguments.

Also please refer to other desktop platform implementations such as [`wWinMain`](https://github.com/flutter/flutter/blob/753fee42cfc62b262a0de90a8acf6a1c2534869c/packages/flutter_tools/templates/app_shared/windows.tmpl/runner/main.cpp.tmpl#L25). They pass the arguments as they are without storing them to somewhere else.

embedding/cpp/flutter_app.cc
swift-kim: Of course we can implement it, if we need.

WonyoungChoi:
> @JSUYA Actually I haven't noticed there's the `--dart-entrypoint-args` option in `flutter-tizen run`, but unfortunately the option cannot be supported on Tizen because there's no way to pass the arguments (`debuggingOptions.dartEntrypointArgs`) to the app in the [`TizenDevice.startApp`](https://github.com/flutter-tizen/flutter-tizen/blob/940c43a8a2577f6efe395f565e061ffd21120d10/lib/tizen_device.dart#L322-L469) function. The `app_launcher -e` command (or the corresponding secure command) doesn't support passing extra arguments.

How about using a temp file like `EngineArgs`? It is possible to upload `--dart-entrypoint-args`  value as a temp file and parse it at before calling `main` of dart.

embedding/cpp/flutter_app.cc
bbrto21: I don't think it's necessary right now.

bbrto21:
> Also please refer to other desktop platform implementations such as [`wWinMain`](https://github.com/flutter/flutter/blob/753fee42cfc62b262a0de90a8acf6a1c2534869c/packages/flutter_tools/templates/app_shared/windows.tmpl/runner/main.cpp.tmpl#L25). They pass the arguments as they are without storing them to somewhere else.

Should we do this too?

swift-kim:
@WonyoungChoi That will be quite a bit of work and will add more complexity to our embedding. Let me take a look if you think it's worth.

@bbrto21 Do what? There's nothing more to do.

bbrto21:
> @bbrto21 Do what? There's nothing more to do.


Sorry, I was mistaken that our template didn't have the implementation below :laughing: 
```
return app.Run(argc, argv);
```

JSUYA:
>@JSUYA Actually I haven't noticed there's the --dart-entrypoint-args option in flutter-tizen run, but unfortunately the option cannot be supported on Tizen because there's no way to pass the arguments (debuggingOptions.dartEntrypointArgs) to the app in the [TizenDevice.startApp](https://github.com/flutter-tizen/flutter-tizen/blob/940c43a8a2577f6efe395f565e061ffd21120d10/lib/tizen_device.dart#L322-L469) function. The app_launcher -e command (or the corresponding secure command) doesn't support passing extra arguments.

@swift-kim 
I looked up an implementation of flutter with regards to --dart-entrypoint-args and created https://github.com/flutter-tizen/flutter-tizen/pull/389. please check this
(Since this PR is the same implementation as WinMain, no more modifications.)

JSUYA (approved):

swift-kim:
Decided to implement only `--dart-entrypoint-args` after internal discussion. Closing.

PR #388, opened 2022-07-04
Remove NotifyAppIsInactive and NotifyAppIsDetached

swift-kim:
- Remove unused `NotifyAppIsInactive` and `NotifyAppIsDetached` from `FlutterEngine`.
- Switch access modifiers of some protected members of `FlutterApp`/`FlutterServiceApp` to private.
- Update stale documentation.

### DIFF
--- a/embedding/cpp/flutter_engine.cc
+++ b/embedding/cpp/flutter_engine.cc
@@ -14,6 +14,7 @@ namespace {
 
 std::vector<std::string> ParseEngineArgs() {
   std::vector<std::string> engine_args;
+
   char* app_id;
   if (app_get_id(&app_id) != 0) {
     TizenLog::Warn("The app ID is not found.");
--- a/embedding/cpp/flutter_engine.cc
+++ b/embedding/cpp/flutter_engine.cc
@@ -131,21 +131,9 @@ void FlutterEngine::NotifyAppIsResumed() {
   }
 }
 
-void FlutterEngine::NotifyAppIsInactive() {
-  if (engine_) {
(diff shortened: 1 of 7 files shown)


bbrto21:
> * Remove unused `NotifyAppIsInactive` and `NotifyAppIsDetached` from `FlutterEngine`.

Please explain the reason for the removal in more detail.

bbrto21:
> * Switch access modifiers of some protected members of `FlutterApp`/`FlutterServiceApp` to private.

`FlutterApp/FlutterServiceApp` are designed as the base class for user application. so protected members are open to users just like public members. In this situation, don't we need a deprecation procedure?
(Of course, I guess there are fewer users who use protected members directly, so I don't think there's a problem.)

swift-kim:
> * Remove unused `NotifyAppIsInactive` and `NotifyAppIsDetached` from `FlutterEngine`.

As you can see in https://api.flutter.dev/flutter/dart-ui/AppLifecycleState.html, the "inactive" or "detached" state is not applicable for the Tizen platform. So they aren't currently used and are not likely to be used. We can add them back when we need.

> * Switch access modifiers of some protected members of `FlutterApp`/`FlutterServiceApp` to private.

As you can remember we've been removing "protected" members without any notice or deprecation process. Also the affected members are not experimental (e.g. `window_offset_x`) so it's unlikely they are used by any existing application.

swift-kim:
@bbrto21 As we've talked in offline, the `inactive` state doesn't seem to be a prerequisite for the `pause` state. The states are mutually independent.

bbrto21 (approved):
LGTM

PR #390, opened 2022-07-05
Add the C# implementation of FlutterEngine

swift-kim:
- Add the `FlutterEngine` class and update `FlutterApplication` and `FlutterServiceApplication` accordingly.
- Move the `ParseEngineArgs()` code to `FlutterEngine` and remove `Utils.cs`.
- Remove `is_running_` from the C++ embedding's `FlutterEngine`. The value doesn't represent the engine's running state correctly if the engine has been started by the view (without calling `Run`).


embedding/csharp/Tizen.Flutter.Embedding/FlutterApplication.cs
```diff
@@ -93,28 +87,10 @@ protected override void OnCreate()
...
-                    dart_entrypoint_argc = entrypointArgs.Length,
-                    dart_entrypoint_argv = entrypointArgs.Handle,
-                };
-
-                Engine = FlutterDesktopEngineCreate(ref engineProperties);
-                if (Engine.IsInvalid)
-                {
-                    throw new Exception("Could not create a Flutter engine.");
-                }
+                throw new Exception("Could not create a Flutter engine.");
```
JSUYA, line 93: Inside FlutterEngine() , it already checks `Engine.IsInvalid` and throws an exception with the same message.
I think we can reduce the exception throw or send a different message.

swift-kim: Good point! I changed the constructor not to throw an exception. How do you like? Now it's the user's responsibility to check if the created `FlutterEngine` is valid or not.

JSUYA (commented):
LGTM :)

JSUYA (approved):

PR #392, opened 2022-07-11
Copy plugin resource files to app TPK

swift-kim:
Copy a plugin's `res` directory to `res/[plugin_name]` in the output TPK during app build.

This feature has been requested by @xiaowei-guan. He wants to selectively dlopen a shared library (bundled with a plugin) depending on the current platform version at runtime.

**Known issue:** The incremental build may fail (plugins do not rebuild) if a new lib or res file has been added after an initial build without modification to any other source file. Possible solutions include:

1. Always force rebuild plugins (this affects the overall build time), or
2. Support the "sharedLib" plugin type (which has been deprecated as of https://github.com/flutter-tizen/flutter-tizen/pull/98) and let the user explicitly specify lib names in `project_def.prop`. (This is a partial solution since res files are not listed by `project_def.prop`.)

### DIFF
--- a/lib/build_targets/package.dart
+++ b/lib/build_targets/package.dart
@@ -120,6 +120,10 @@ class DotnetTpk extends TizenPackage {
     if (pluginsLib.existsSync()) {
       pluginsLib.copySync(libDir.childFile(pluginsLib.basename).path);
     }
+    final Directory pluginsResDir = pluginsDir.childDirectory('res');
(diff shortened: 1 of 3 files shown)


xiaowei-guan (approved):
Looks good to me, thank you~

bbrto21:
I think you should remove the `res` dir in the build root before copying each plugins's res dir.

bbrto21 (approved):

PR #393, opened 2022-07-12
Use renderer type for engine

JSUYA:
Set the renderer type before creating the engine.
The window is selected as EcoreWl2 or ElmWin
according to the renderer type.

This commit requires engine side commits.
(https://github.com/flutter-tizen/engine/pull/312)


JSUYA:
I think `FlutterDesktopViewCreateFromNewEcoreWl2Window` and `FlutterDesktopViewCreateFromNewElmWindow` can be changed with one API. This can be done with additional work.

JSUYA:
> I think `FlutterDesktopViewCreateFromNewEcoreWl2Window` and `FlutterDesktopViewCreateFromNewElmWindow` can be changed with one API. This can be done with additional work.

This is a problem that is implemented incorrectly in the engine. I will fix it soon

embedding/csharp/Tizen.Flutter.Embedding/Interop/Interop.cs
```diff
@@ -24,6 +30,7 @@ public struct FlutterDesktopWindowProperties
             public bool focusable;
             [MarshalAs(UnmanagedType.U1)]
             public bool top_level;
+            public int renderer_type;
```
swift-kim, line 33: Casting should not be necessary.
```suggestion
            public FlutterDesktopRendererType renderer_type;
```

embedding/csharp/Tizen.Flutter.Embedding/FlutterApplication.cs
```diff
@@ -51,6 +51,11 @@ public class FlutterApplication : CoreUIApplication, IPluginRegistry
         /// </summary>
         protected bool IsTopLevel { get; set; } = false;
 
+        /// <summary>
+        /// The renderer type of Flutter engine.
```
swift-kim, line 55: The description should match with the C++ embedding's.

embedding/csharp/Tizen.Flutter.Embedding/FlutterApplication.cs
```diff
@@ -51,6 +51,11 @@ public class FlutterApplication : CoreUIApplication, IPluginRegistry
...
+        /// </summary>
+        protected FlutterDesktopRendererType RendererType { get; set; } = FlutterDesktopRendererType.kEGL;
```
swift-kim, line 57: What if the selected type is not supported by the current profile? (i.e. wearable)

How do we automatically select the `kEvasGL` type for wearable apps?

embedding/cpp/elm_flutter_view.cc
```diff
@@ -16,7 +16,7 @@ bool ElmFlutterView::RunEngine() {
   }
 
   if (!engine_) {
-    engine_ = FlutterEngine::Create();
+    engine_ = FlutterEngine::Create("", std::vector<std::string>());
```
swift-kim, line 19: Is this change intentional?

embedding/csharp/Tizen.Flutter.Embedding/Interop/Interop.cs
JSUYA: done

embedding/csharp/Tizen.Flutter.Embedding/FlutterApplication.cs
JSUYA: done

embedding/csharp/Tizen.Flutter.Embedding/FlutterApplication.cs
JSUYA: I added profile check macro for this.

embedding/cpp/elm_flutter_view.cc
JSUYA: my mistake.. thank you :)

bbrto21:
Could you please validate a service app? a service app goes into the crash with this PR and https://github.com/flutter-tizen/engine/pull/312
https://github.com/flutter-tizen/flutter-tizen/wiki/Application-types#service-app

embedding/csharp/Tizen.Flutter.Embedding/FlutterApplication.cs
```diff
@@ -51,6 +51,15 @@ public class FlutterApplication : CoreUIApplication, IPluginRegistry
         /// </summary>
         protected bool IsTopLevel { get; set; } = false;
 
+        /// <summary>
+        /// The renderer type of the engine. Defaults to kEGL. If the profile is wearable, defaults to kEvasGL.
+        /// </summary>
+#if WEARABLE_PROFILE
+        protected FlutterDesktopRendererType RendererType { get; set; } = FlutterDesktopRendererType.kEvasGL;
+#else
+        protected FlutterDesktopRendererType RendererType { get; set; } = FlutterDesktopRendererType.kEGL;
```
swift-kim, line 60: Avoid declaring the same property twice. You may either move the location of `#if` or assign the value inside the class constructor.

Also the `WEARABLE_PROFILE` macro is not defined.

JSUYA: > Avoid declaring the same property twice. You may either move the location of #if or assign the value inside the class constructor.

I will move it to constructor.

> Also the WEARABLE_PROFILE macro is not defined.

https://github.com/flutter-tizen/flutter-tizen/blob/592bb17f6b91eb592485389ba738295e98e7b261/lib/build_targets/package.dart#L364

I thought I could use a macro with this predefine.
Is there any way to get the profile in flutter-tizen?

swift-kim: Yeah, the macro is defined by the C++ compiler but not by the .NET compiler. The .NET equivalent is to
- pass the `/p:DefineConstants=WEARABLE_PROFILE` property to the `dotnet build` command as an argument (you need to modify `package.dart`), or
- write the property to the `Runner.csproj.user` file (you need to modify `updateDotnetUserProjectFile` of `tizen_project.dart`).

I'd like to avoid the both but if we really have to choose one, I would prefer the latter.

embedding/cpp/include/flutter_app.h
```diff
@@ -81,6 +81,16 @@ class FlutterApp : public flutter::PluginRegistry {
   // must be added to tizen-manifest.xml file.
   bool is_top_level_ = false;
 
+  // The renderer type of the engine.
+  //
+  // Defaults to kEGL. If the profile is wearable, defaults to kEvasGL.
+#ifdef WEARABLE_PROFILE
+  FlutterDesktopRendererType renderer_type_ =
```
swift-kim, line 88: The type `FlutterDesktopRendererType` is not expected to be exposed through the public interface of the embedding. I'll talk to you offline in detail.

embedding/csharp/Tizen.Flutter.Embedding/FlutterApplication.cs
JSUYA: Thank you for your advice.
I tried the second of the ways you suggested, but it didn't work the way I thought it would
I added the below code to tizen_project.dart and tested it, but the WEARABLE_PROFILE define was not found in C#.

```dart
 final Iterable<XmlElement> elementsDefineConstants =
      document.findAllElements('DefineConstants');
  if (elementsDefineConstants.isEmpty) {
    final XmlBuilder builder = XmlBuilder();
    builder.element('PropertyGroup', nest: () {
      builder.element(
        'DefineConstants',
        nest: 'WEARABLE_PROFILE',
      );
    });
    document.rootElement.children.add(builder.buildFragment());
  } else {
    for (final XmlElement element in elementsDefineConstants) {
      element.innerText = 'WEARABLE_PROFILE';
    }
  }
```
Runner.csproj.user
```csproj
<?xml version="1.0" encoding="utf-8"?>
<Project ToolsVersion="Current" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <PropertyGroup>
    <FlutterEmbeddingPath>/home/junsu/dev/os/f-project/flutter-tizen/embedding/csharp/Tizen.Flutter.Embedding/Tizen.Flutter.Embedding.csproj</FlutterEmbeddingPath>
  </PropertyGroup>
  <PropertyGroup>
    <DefineConstants>WEARABLE_PROFILE</DefineConstants>
  </PropertyGroup>
</Project>
```

How about adding it to package.dart as the first of your suggestions?
I think this way is better because it can direct access(?) to buildInfo.deviceProfile.
```diff
--- a/lib/build_targets/package.dart
+++ b/lib/build_targets/package.dart
@@ -144,6 +144,7 @@ class DotnetTpk extends TizenPackage {
       if (buildMode.isPrecompiled) 'Release' else 'Debug',
       '-o',
       '${outputDir.path}/', // The trailing '/' is needed.
+      '/p:DefineConstants=${buildInfo.deviceProfile.toUpperCase()}_PROFILE',
       tizenProject.editableDirectory.path,
     ]);
     if (result.exitCode != 0) {

```

swift-kim: Alright. Only the first option seems to be feasible.

embedding/cpp/include/flutter_app.h
```diff
@@ -15,10 +15,22 @@
...
 // The app base class for headed Flutter execution.
 class FlutterApp : public flutter::PluginRegistry {
  public:
-  explicit FlutterApp() {}
+  explicit FlutterApp()
+#ifdef WEARABLE_PROFILE
+      : renderer_type_(FlutterRendererType::kEvasGL)
+#endif
+  {
+  }
```
swift-kim, line 33: nit: This might look cleaner. (Would there be any logical difference?)
```suggestion
  explicit FlutterApp() {
#ifdef WEARABLE_PROFILE
    renderer_type_ = FlutterRendererType::kEvasGL;
#endif
  }
```

embedding/csharp/Tizen.Flutter.Embedding/FlutterApplication.cs
```diff
@@ -10,11 +10,36 @@
...
+    public enum FlutterRendererType
+    {
+        /// <summary>
+        /// The renderer based on EvasGL.
+        /// </summary>
+        EvasGL = 0,
+        /// <summary>
+        /// The renderer based on EGL.
+        /// </summary>
+        EGL
```
swift-kim, line 25: nit: Just for consistency.
```suggestion
        EGL,
```

lib/build_targets/package.dart
```diff
@@ -145,6 +145,7 @@ class DotnetTpk extends TizenPackage {
       '-o',
       '${outputDir.path}/', // The trailing '/' is needed.
       tizenProject.editableDirectory.path,
+      '/p:DefineConstants=${buildInfo.deviceProfile.toUpperCase()}_PROFILE',
```
swift-kim, line 148: Hmm, I'm not sure which way is the standard but wouldn't it be better to make the option come before the argument (directory path)?

lib/build_targets/package.dart
```diff
@@ -285,7 +286,10 @@ class NativeTpk extends TizenPackage {
       embeddingDir.path,
       configuration: buildConfig,
       arch: getTizenCliArch(buildInfo.targetArch),
-      extraOptions: <String>['-fPIC'],
+      extraOptions: <String>[
+        '-fPIC',
+        '-D${buildInfo.deviceProfile.toUpperCase()}_PROFILE'
```
swift-kim, line 291: Please use

```dart
        predefines: <String>[
          '${buildInfo.deviceProfile.toUpperCase()}_PROFILE',
        ],
```

for consistency. (Please refer to `plugins.dart` for the exact order between the arguments.)

test/general/build_targets/package_test.dart
```diff
@@ -93,6 +95,7 @@ void main() {
           '-o',
           '${outputDir.path}/tpk/',
           '${projectDir.path}/tizen',
+          '/p:DefineConstants=${deviceProfile.toUpperCase()}_PROFILE',
```
swift-kim, line 98: Using inlined values makes the test more robust because you can check if the value is parsed and passed to the builder correctly as well.
```suggestion
          '/p:DefineConstants=COMMON_PROFILE',
```

swift-kim (commented):
Only minor nitpicks.

Could you also change the file mode of `FlutterEngine.cs` to 644? (It's accidentally 755 right now.)

embedding/cpp/include/flutter_app.h
swift-kim: > Should these be fixed later as well?

Well, I think the cases you mentioned are implemented in that way on purpose (to directly access underlying native pointers) and thus won't need a fix. However, let me take a look again.

JSUYA:
> Could you please validate a service app? a service app goes into the crash with this PR and [flutter-tizen/engine#312](https://github.com/flutter-tizen/engine/pull/312) https://github.com/flutter-tizen/flutter-tizen/wiki/Application-types#service-app

I tested on multi-app samples. works fine

embedding/cpp/flutter_app.cc
```diff
@@ -16,6 +16,13 @@ bool FlutterApp::OnCreate() {
     TizenLog::Error("Could not create a Flutter engine.");
     return false;
   }
+#ifdef WEARABLE_PROFILE
+  if (renderer_type_ == FlutterRendererType::kEGL) {
+    TizenLog::Error(
+        "FlutterRendererType.kEGL is not supported by this profile.");
```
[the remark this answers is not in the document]
JSUYA: done

embedding/cpp/include/flutter_app.h
JSUYA:
```cpp
A a = 1;
```
```cpp
A a;
a = 1;
```
I know the difference between these.

JSUYA: I think it doesn't matter now. I like looks cleaner too

embedding/csharp/Tizen.Flutter.Embedding/FlutterApplication.cs
JSUYA: done

lib/build_targets/package.dart
JSUYA: There is no special reason. I was worried that it might affect other flags. I modified according to your comments.

lib/build_targets/package.dart
JSUYA: oh thanks. i just understanded this out

test/general/build_targets/package_test.dart
JSUYA: done

embedding/cpp/include/flutter_app.h
swift-kim: `IPluginRegistry` is a C# version of C++'s [`PluginRegistry`](https://github.com/flutter-tizen/engine/blob/flutter-3.0.0-tizen/shell/platform/common/client_wrapper/include/flutter/plugin_registry.h), and its `GetRegistrarForPlugin` method is expected to return a raw `FlutterDesktopPluginRegistrar` handle.

That said, we can still provide a higher level API of `FlutterDesktopPluginRegistrar` in the C# side which provides the same functionality as C++'s [`PluginRegistrar`](https://github.com/flutter-tizen/engine/blob/flutter-3.0.0-tizen/shell/platform/common/client_wrapper/include/flutter/plugin_registrar.h) class. However, even if it's implemented, there will be no use for it.

swift-kim: Also I was thinking of introducing a new public interface for `FlutterApp` that can be used to configure all (currently experimental) window-related properties (such as `window_offset_x_` and `is_window_transparent_`) and also the renderer type in this PR at once using a single struct. I don't have a concrete plan yet. You may think about it if interested (maybe later).

bin/internal/engine.version
```diff
@@ -1 +1 @@
-3.0.0-06ddbce
+3.0.0-6c699b2
```
swift-kim, line 1: Could you update the hash to the latest (https://github.com/flutter-tizen/engine/commit/1ca586769cc68909c16dc2ccacbd05b8ffeaf1df)?

swift-kim (approved):

bin/internal/engine.version
JSUYA: done

PR #394, opened 2022-07-13
Use native UI toolkit APIs for manipulation of view

bbrto21:
* If you can get the native handle from the view,
  and the manipulation for the handle is open to you,
  the view should respond appropriately to the manipulation.
  therefore, we don't need a view's resize API.

Signed-off-by: Boram Bae <boram21.bae@samsung.com>

### DIFF
--- a/embedding/cpp/elm_flutter_view.cc
+++ b/embedding/cpp/elm_flutter_view.cc
@@ -51,12 +51,7 @@ void ElmFlutterView::SetEngine(std::unique_ptr<FlutterEngine> engine) {
 void ElmFlutterView::Resize(int32_t width, int32_t height) {
   assert(IsRunning());
 
-  int32_t view_width = width, view_height = height;
-  evas_object_geometry_get(evas_object_, nullptr, nullptr, &view_width,
-                           &view_height);
-  if (view_width != width || view_height != height) {
-    FlutterDesktopViewResize(view_, width, height);
-  }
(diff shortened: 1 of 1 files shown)


bbrto21:
@JSUYA How about this draft? 
if you agree this then we can remove `FlutterDestkopViewResize` from Embedder APIs.

embedding/cpp/elm_flutter_view.cc
```diff
@@ -51,12 +51,7 @@ void ElmFlutterView::SetEngine(std::unique_ptr<FlutterEngine> engine) {
 void ElmFlutterView::Resize(int32_t width, int32_t height) {
   assert(IsRunning());
 
-  int32_t view_width = width, view_height = height;
-  evas_object_geometry_get(evas_object_, nullptr, nullptr, &view_width,
-                           &view_height);
-  if (view_width != width || view_height != height) {
-    FlutterDesktopViewResize(view_, width, height);
-  }
+  evas_object_resize(evas_object_, width, height);
```
JSUYA, line 54: I think this is not appropriate because it depends on resize event callback of view_.
I think this Resize method is an API for view_. I don't think we need to manipulate evas_object_ directly in this Resize method.

bbrto21: All right, your opinion makes sense, and this PR doesn't make any functional changes, so I'll close this PR.

Issue #397, opened 2022-07-14
Support Flutter-Tizen as a module


bbrto21:
As a first step to provide a way to easily porting Flutter-Tizen to an existing Tizen project, we can consider supporting the below instruction set.

```
cd ~/existing_tizen_project_path/
flutter-tizen create -t module ./some-flutter-tizen-module
flutter-tizen build sharedLib
```

In order to provide flutter-tizen as a sharedLib, the module should not have an entry point in the native area, and the public header of the `embedding` should be exposed in the built sharedlib.

Intuitively, you can imagine the build output like the one below.
ls ~/existing_tizen_project_path/some-flutter-tizen-module/build
```
build/
└── tizen
    └── sharedLib
         ├── include
         │   ├── HEADER_FILES_OF_EMBEDDING
         │   ├──  ......
         ├── lib
         │   ├── libapp.so
         │   ├── libflutter_engine.so
         │   └── libflutter_tizen_mobile.so
         ├── res
         │   ├── flutter_assets
         │   │   ├── AssetManifest.json
         │   │   ├── FontManifest.json
         │   │   ├── fonts
         │   │   │   └── MaterialIcons-Regular.otf
         │   │   ├── NOTICES.Z
         │   │   └── packages
         │   │       └── cupertino_icons
         │   │           └── assets
         │   │               └── CupertinoIcons.ttf
         │   └── icudtl.dat
         └── libsome-flutter-tizen-module.so
```
After that, the user can manually set dependency and use a module in the existing tizen project.
Oh, of course, the user has to write the program manually using the set of public API provided by the embedding headers.
I think there are many loopholes in this draft. Let's discuss it together.

WonyoungChoi:
So... do users have to copy the embedding headers and flutter related .so files directly to existing projects?

bbrto21:
I imagined the directory would be organized like this after execution of `module` command

`ls ~/existing_tizen_project_path/some-flutter-tizen-module`
```
some-flutter-tizen-module/
├── lib
│   └── main.dart
├── pubspec.yaml
├── README.md
└── test
    └── widget_test.dart
```
and after build a module, since it is already located under the existing Tizen project, I thought the user could set up dependencies immediately without copying. of course, users must change the project configuration to include the built modules into the app package(tpk)

As you mentioned, I think If the user created the module in some place other than the existing Tizen project, then user can copy the sharedLib directory to use it.

[...]

PR #402, opened 2022-07-26
Support creating a module type project

swift-kim:
Part of https://github.com/flutter-tizen/flutter-tizen/issues/397.

- Add module templates (for C# and C++) and relevant patch files.
- Unlike other types of projects, a module project is not created by the `create` command itself. (Thus there's no need to "copy" the template to `flutter_tools/templates`.) Instead, the project is created by `TizenProject.ensureReadyForPlatformSpecificTooling` only if the target (either `tizen` or `.tizen`) doesn't already exist.
- You will see the following line at the end of the created project's `pubspec.yaml`. The information is used by `TizenProject.ensureReadyForPlatformSpecificTooling` for choosing a right template.<p>
  ```yaml
  flutter:
    module:
      tizenLanguage: cpp
  ```
- Change the location of the `flutter` directory of a multi app from `tizen/` (editableDirectory) to `tizen/ui/` (hostAppRoot).


lib/commands/create.dart
```diff
@@ -80,12 +80,14 @@ class TizenCreateCommand extends CreateCommand {
...
-      return stringArg('template') == 'plugin';
-    } else if (projectDir.existsSync() && projectDir.listSync().isNotEmpty) {
-      return determineTemplateType() == FlutterProjectType.plugin;
+      template = stringToProjectType(stringArg('template'));
     }
-    return false;
+    if (projectDir.existsSync() && projectDir.listSync().isNotEmpty) {
+      template = determineTemplateType();
+    }
+    return template == FlutterProjectType.plugin;
```
swift-kim, line 90: I just refactored this part. There's no important logic change.

lib/tizen_plugins.dart
```diff
@@ -291,9 +291,6 @@ Future<void> ensureReadyForTizenTooling(FlutterProject project) async {
     return;
   }
   final TizenProject tizenProject = TizenProject.fromFlutter(project);
-  if (!tizenProject.existsSync()) {
-    return;
-  }
```
swift-kim, line 296: Removing this check should be safe, because the check is also performed by `injectTizenPlugins`.

lib/tizen_project.dart
```diff
@@ -21,15 +26,39 @@ class TizenProject extends FlutterProjectPlatform {
 
   final FlutterProject parent;
 
+  /// See: [FlutterManifest] in `flutter_manifest.xml`
+  late final Map<String, Object?> _parentPubspec = () {
```
swift-kim, line 30: We can't use `parent.manifest` directly and should parse the file again because `FlutterManifest._descriptor` is private.

lib/tizen_project.dart
```diff
@@ -21,15 +26,39 @@ class TizenProject extends FlutterProjectPlatform {
...
   Directory get editableDirectory => parent.directory.childDirectory('tizen');
+  Directory get _ephemeralModuleDirectory =>
+      parent.directory.childDirectory('.tizen');
+
+  Directory get hostAppRoot {
+    if (!parent.isModule || editableDirectory.existsSync()) {
+      return isMultiApp ? uiAppDirectory : editableDirectory;
+    }
+    return _ephemeralModuleDirectory;
+  }
```
swift-kim, line 56: `hostAppRoot` may refer to either
- `.tizen/` (ephemeralModuleDirectory)
- `tizen/` (editableDirectory)
- `tizen/ui/` (uiAppDirectory)

according to the context.

lib/tizen_project.dart
```diff
@@ -38,65 +67,114 @@ class TizenProject extends FlutterProjectPlatform {
...
-      (isMultiApp
-          ? uiAppDirectory.childFile('project_def.prop')
-          : editableDirectory.childFile('project_def.prop'));
+  File get manifestFile => hostAppRoot.childFile('tizen-manifest.xml');
 
   @override
-  bool existsSync() => editableDirectory.existsSync();
+  bool existsSync() => hostAppRoot.existsSync();
+
+  File? get projectFile {
```
swift-kim, line 92: Now this property has a nullable return type.

bbrto21 (approved):
Thanks!

PR #407, opened 2022-08-03
Support sharedLib type native plugins

swift-kim:
Resolves https://github.com/flutter-tizen/plugins/issues/417. This is also an alternative solution of https://github.com/flutter-tizen/flutter-tizen/pull/392.

- Support both "staticLib" and "sharedLib" type plugins.
- "sharedLib" type plugins are built into individual shared objects (`lib{plugin_name}.so`). They can specify arbitrary linker options in their `project_def.prop`. In case of `webview_flutter_tizen`, for example:<p>
  ```ini
  USER_LIBS = pthread lightweight-web-engine.flutter tuv
  USER_LIB_DIRS = lib/${BUILD_ARCH}
  USER_LFLAGS =
  ```
- Let `TizenPlugins` depend on `NativeEmbedding` (whose build output contains compiled cpp_client_wrapper).
- Add missing input dependencies to `NativeEmbedding`.
- Change the output directory of `libflutter_plugins.so` from `{BUILD_ROOT}/tizen_plugins/` to `{BUILD_ROOT}/tizen_plugins/lib/`.

Deprecation notice: The "automatic user lib linking" support for "staticLib" type plugins will be removed in the future.

### DIFF
--- a/lib/build_targets/embedding.dart
+++ b/lib/build_targets/embedding.dart
@@ -29,6 +29,7 @@ class NativeEmbedding extends Target {
   @override
   List<Source> get inputs => const <Source>[
         Source.pattern('{FLUTTER_ROOT}/../lib/build_targets/embedding.dart'),
+        Source.pattern('{FLUTTER_ROOT}/../lib/tizen_sdk.dart'),
       ];
 
   @override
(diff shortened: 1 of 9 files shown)


HakkyuKim (approved):
LGTM

Issue #351, opened 2022-03-29
Support FlutterView


bbrto21:
There are internal requirements for `FlutterView`.

The main purpose is to provide some interface for each language(`C#`/`C++`) that works as a UI component of Tizen rather than a single page FlutterApp. (see : https://docs.flutter.dev/development/add-to-app)
Although specific requirements have not yet been communicated, we may need to support both. however, our top priority is to provide Flutter-tizen as a [NUI](https://docs.tizen.org/application/dotnet/guides/nui/overview/) component.

I think there are two ways to achieve this. the first way is to provide it by implementing it in embeddings, and the other way is to provide it by implementing in Tizen platform level after including flutter-tizen to Tizen platform like the webview provided in Tizen UI Toolkit.  each way has distinct pros and cons I think but I'm approaching it the first way.

Several issues have already been opened and reviewed, but it seems to me that they all serve the same purpose.
so I will close these issues in favor of managing them as a single issue. (@JSUYA agreed to this)
 - https://github.com/flutter-tizen/flutter-tizen/issues/325
 - https://github.com/flutter-tizen/engine/issues/253
 - https://github.com/flutter-tizen/engine/issues/233

Here's a rough to-do list.
(Please feel free to add a specific to do item. honestly, I still don't know exactly what to do :sweat_smile:)
 - [ ] Support FlutterView on C#
     - [ ] Support FlutterView working on [NUI](https://docs.tizen.org/application/dotnet/api/TizenFX/API9/api/Tizen.NUI.html)
     - [ ] Support FlutterView working on  [ElmSharp](https://docs.tizen.org/application/dotnet/api/TizenFX/API9/api/ElmSharp.html)
 - [ ] Support FlutterView on C++
     - [ ] Support FlutterView working on  [Elementary](https://docs.tizen.org/application/native/api/mobile/6.5/group__Elementary.html)

bbrto21:
Things that have been revealed about NUI
- NUI is based on dali2 from Tizen 6.0 and later. and currently, we cannot use Dali 2 in our engine build configuration. see [this](https://github.com/flutter-tizen/engine/issues/253)
- [NUIApplication](https://docs.tizen.org/application/dotnet/api/TizenFX/API9/api/Tizen.NUI.NUIApplication.html) can be used to provide a embedding. Window of NUIAapplication is based on wl2 and can obtain a native handle from Window instance, so we don't have to create a wl2 window ourselves on engine side for single page application scenario.
[...]

PR #416, opened 2022-08-08
Add the C# implementation of ElmFlutterView

swift-kim:
Contributes to https://github.com/flutter-tizen/flutter-tizen/issues/351.

Changes:
- Add the C# version of C++'s `ElmFlutterView` class.
- Add `IsRunning` to `FlutterApplication` and `FlutterServiceApplication` and replace `Engine.IsValid` with `IsRunning` (for overall consistency).
- Add missing documentation.
- Minor fixes and formatting.

Usage:
```csharp
var flutterView = new ElmFlutterView(box);
if (flutterView.RunEngine())
{
    GeneratedPluginRegistrant.RegisterPlugins(flutterView);
    box.PackEnd(flutterView.EvasObject);
}

...

if (flutterView.IsRunning)
{
    // The view must be explicitly destroyed by user.
    flutterView.Destroy();
}
```

### DIFF
--- a/embedding/cpp/elm_flutter_view.cc
+++ b/embedding/cpp/elm_flutter_view.cc
@@ -12,8 +12,6 @@
 ElmFlutterView::~ElmFlutterView() {
   if (view_) {
     FlutterDesktopViewDestroy(view_);
-    engine_ = nullptr;
-    view_ = nullptr;
   }
 }
 
--- a/embedding/cpp/elm_flutter_view.cc
+++ b/embedding/cpp/elm_flutter_view.cc
@@ -12,6 +12,7 @@
 ElmFlutterView::~ElmFlutterView() {
   if (view_) {
     FlutterDesktopViewDestroy(view_);
+    view_ = nullptr;
   }
 }
 
--- a/embedding/cpp/elm_flutter_view.cc
+++ b/embedding/cpp/elm_flutter_view.cc
@@ -12,6 +12,7 @@
 ElmFlutterView::~ElmFlutterView() {
   if (view_) {
     FlutterDesktopViewDestroy(view_);
+    engine_ = nullptr;
     view_ = nullptr;
   }
 }
--- a/embedding/csharp/Tizen.Flutter.Embedding/ElmFlutterView.cs
+++ b/embedding/csharp/Tizen.Flutter.Embedding/ElmFlutterView.cs
@@ -17,10 +17,6 @@ namespace Tizen.Flutter.Embedding
 
         protected override IntPtr CreateHandle(EvasObject parent)
         {
-            if (Handle == IntPtr.Zero)
-            {
-                throw new InvalidOperationException("The handle cannot be created again.");
-            }
             return Handle;
         }
     }
@@ -60,6 +56,9 @@ namespace Tizen.Flutter.Embedding
         /// </summary>
         private int InitialHeight { get; set; } = 0;
 
+        /// <summary>
+        /// Whether the view is running.
+        /// </summary>
         public bool IsRunning => !View.IsInvalid;
 
         /// <summary>
@@ -100,6 +99,9 @@ namespace Tizen.Flutter.Embedding
             InitialHeight = initialHeight;
         }
 
+        /// <summary>
+        /// Starts running the view with the associated engine, creating if not set.
+        /// </summary>
         public bool RunEngine()
         {
             if (Parent == null)
@@ -141,11 +143,17 @@ namespace Tizen.Flutter.Embedding
             return true;
         }
 
+        /// <summary>
+        /// Sets an engine associated with this view.
+        /// </summary>
         public void SetEngine(FlutterEngine engine)
         {
             Engine = engine;
         }
 
+        /// <summary>
+        /// Terminates the running view and the associated engine.
+        /// </summary>
         public void Destroy()
         {
             if (IsRunning)
@@ -155,6 +163,9 @@ namespace Tizen.Flutter.Embedding
             }
         }
 
+        /// <summary>
+        /// Resizes the view.
+        /// </summary>
         public void Resize(int width, int height)
         {
             Debug.Assert(IsRunning);
@@ -165,10 +176,6 @@ namespace Tizen.Flutter.Embedding
             }
         }
 
-        /// <summary>
-        /// Returns the plugin registrar handle for the plugin with the given name.
-        /// The name must be unique across the application.
-        /// </summary>
         public FlutterDesktopPluginRegistrar GetRegistrarForPlugin(string pluginName)
         {
             if (Engine.IsValid)
--- a/embedding/csharp/Tizen.Flutter.Embedding/ElmFlutterView.cs
+++ b/embedding/csharp/Tizen.Flutter.Embedding/ElmFlutterView.cs
@@ -5,11 +5,9 @@ using static Tizen.Flutter.Embedding.Interop;
 
 namespace Tizen.Flutter.Embedding
 {
-    // TODO: Revisit EvasObject.cs implementation.
-    // TODO: consider renaming EvasObjectImpl.
-    // TODO: Add documentation.
-    // TODO(another PR): Format the C# code.
-    // TODO: Runner.csproj: nuget or ProjectReference.
+    /// <summary>
+    /// Represents an <see cref="EvasObject"/> instance created by the embedder.
+    /// </summary>
     class EvasObjectImpl : EvasObject
     {
         public EvasObjectImpl(EvasObject parent, IntPtr handle) : base(parent)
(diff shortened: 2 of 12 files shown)


embedding/cpp/elm_flutter_view.cc
```diff
@@ -12,6 +12,8 @@
 ElmFlutterView::~ElmFlutterView() {
   if (view_) {
     FlutterDesktopViewDestroy(view_);
+    engine_ = nullptr;
```
JSUYA, line 15: Is this a necessary? As ElmFlutterView is deleted, engine_ will deleted.

embedding/csharp/Tizen.Flutter.Embedding/ElmFlutterView.cs
```diff
@@ -0,0 +1,189 @@
...
+        /// </summary>
+        public void SetEngine(FlutterEngine engine)
+        {
+            Engine = engine;
+        }
+
+        /// <summary>
+        /// Terminates the running view and the associated engine.
+        /// </summary>
+        public void Destroy()
```
JSUYA, line 157: Is a separate Destroy() necessary? Can you handle it in ElmFlutterView's destructor?

JSUYA (commented):
Nice pr :). I left a few comments.

embedding/csharp/Tizen.Flutter.Embedding/ElmFlutterView.cs
WonyoungChoi, line 157: How about using IDisposable pattern?

embedding/cpp/elm_flutter_view.cc
swift-kim: You're right. I was just trying to make this consistent with `ElmFlutterView.Destroy` (C#) and

https://github.com/flutter-tizen/flutter-tizen/blob/8d6f09c254575b3e66b7d070e78adaec9fa18ed3/embedding/cpp/flutter_app.cc#L57-L63

and didn't notice that it was within the class destructor.

embedding/csharp/Tizen.Flutter.Embedding/ElmFlutterView.cs
swift-kim: The .NET runtime's GC runs at an uncertain point of time during app execution so it didn't seem a good practice to use the destructor to release native resources.

swift-kim: Using the IDisposable pattern in this context didn't make very much sense to me. My concerns were:

- `ElmFlutterView` will (and should) never be used along with `using`.
- The user will not feel natural about the API `view.Dispose()` because the user has no knowledge of whether the view is associated with any unmanaged resources. On the contrary, the name `view.Destroy()` is consistent with `FlutterDesktopViewDestroy`.
- `ElmFlutterView` is not a direct wrapper of the native Flutter view.

embedding/csharp/Tizen.Flutter.Embedding/ElmFlutterView.cs
```diff
@@ -0,0 +1,189 @@
+﻿using System;
```
JSUYA, line 1: No License text required?

JSUYA (approved):
LGTM
+ Could you share some simple app code to test the ElmSharp?

embedding/csharp/Tizen.Flutter.Embedding/ElmFlutterView.cs
swift-kim: I forgot. Thanks.

bbrto21 (approved):

embedding/csharp/Tizen.Flutter.Embedding/ElmFlutterView.cs
WonyoungChoi: What if the `view` object disappears by GC without Destroy () call?

swift-kim: That's an undefined behavior but the view will be left running (even if the parent EvasObject disappears) and no C# code will be able to access the view. I think this is better than the view being abruptly (unintentionally) destroyed by the GC but how do you think?

swift-kim:
@JSUYA The sample code I used for testing can be found at https://github.com/swift-kim/csharp_module/blob/master/tizen/App.cs but I'm really new to EFL and ElmSharp so the code is not very good. The app is intended to be run on a mobile emulator.

embedding/csharp/Tizen.Flutter.Embedding/ElmFlutterView.cs
WonyoungChoi: I think so. we can register the view internally somewhere, but.. That doesn't make any sense, either.

embedding/csharp/Tizen.Flutter.Embedding/ElmFlutterView.cs
```diff
@@ -0,0 +1,193 @@
...
+                return false;
+            }
+
+            return true;
+        }
+
+        /// <summary>
+        /// Sets an engine associated with this view.
+        /// </summary>
+        public void SetEngine(FlutterEngine engine)
```
WonyoungChoi, line 153: Is this setter necessary? If this setter is used after `RunEngine()` is called, it may be a problem. If necessary, why don't we move to the constructor's parameter or RunEngine's optional parameter?

embedding/csharp/Tizen.Flutter.Embedding/ElmFlutterView.cs
```diff
@@ -0,0 +1,193 @@
...
+
+    /// <summary>
+    /// Displays a Flutter screen in a Tizen application.
+    /// </summary>
+    public class ElmFlutterView : IPluginRegistry
+    {
+        /// <summary>
+        /// The Flutter engine instance.
+        /// </summary>
+        private FlutterEngine Engine { get; set; } = null;
```
WonyoungChoi, line 36: How about changing Engine getter to public? I think we need to deliver the Pause, Resume event to Engine.

embedding/csharp/Tizen.Flutter.Embedding/ElmFlutterView.cs
swift-kim: This API was designed to be consistent with C++'s

https://github.com/flutter-tizen/flutter-tizen/blob/8d6f09c254575b3e66b7d070e78adaec9fa18ed3/embedding/cpp/elm_flutter_view.cc#L54-L56

and yes, we can change the API as you said. However, if we do so, the C++ API will need a change as well and transferring the ownership of `std::unique_ptr<FlutterEngine>` will become less explicit:

```cpp
// |engine| can be nullptr.
bool ElmFlutterView::RunEngine(std::unique_ptr<FlutterEngine> engine) {
  // If engine is nullptr, create one.
  ...
}
```

Is this fine? @bbrto21 How do you think?

embedding/csharp/Tizen.Flutter.Embedding/ElmFlutterView.cs
swift-kim: I just followed https://github.com/flutter-tizen/flutter-tizen/blob/8d6f09c254575b3e66b7d070e78adaec9fa18ed3/embedding/cpp/include/elm_flutter_view.h#L46-L48

and notifying the engine of any event will require a dedicated API or something. (You might refer to https://github.com/flutter-tizen/flutter-tizen/pull/378.)

We can change the access modifier to public (or protected) any time later if we need.

embedding/csharp/Tizen.Flutter.Embedding/ElmFlutterView.cs
bbrto21: if the engine can be used as a parameter in any form, there is still a problem that the engine can run before set engine to view. for now I think, it seems better to deal with only the parameters for the dart project rather than using the engine as a parameter.

embedding/csharp/Tizen.Flutter.Embedding/ElmFlutterView.cs
bbrto21: I had some concerns, so I left [it](https://github.com/flutter-tizen/flutter-tizen/pull/378) as draft.
One of the concerns was to provide the engine reference of the view publicly instead of view's methods for life-cycle. 

Which one looks better(provide an engine reference as public or methods for life-cycle)?

embedding/csharp/Tizen.Flutter.Embedding/ElmFlutterView.cs
```diff
@@ -0,0 +1,193 @@
...
+        {
+            Parent = parent;
+            InitialWidth = initialWidth;
+            InitialHeight = initialHeight;
+        }
+
+        /// <summary>
+        /// Starts running the view with the associated engine, creating if not set.
+        /// </summary>
+        public bool RunEngine()
```
WonyoungChoi, line 109: Should we prevent calling RunEngine() twice?
`IsRunning` can be used for it.

swift-kim: Okay. We may need some unit tests as well.

swift-kim:
All issues have been addressed. Please review again.

WonyoungChoi (approved):
LGTM :+1:

JSUYA (approved):

PR #418, opened 2022-08-10
Introduce NUIFlutterView class

JSUYA:
Contributes to #351.

This is a flutter view class available in NUI UIFW.

```
var flutterView = new NUIFlutterView()
{
    Size2D = new Size2D(width, height),
    Position2D = new Position2D(x, y),
);
if (flutterView.RunEngine())
{
    GeneratedPluginRegistrant.RegisterPlugins(flutterView.Engine);
    ParentView.Add(flutterView);
}
...

if (flutterView.IsRunning)
{
    flutterView.Destroy();
}

```

releated PR : https://github.com/flutter-tizen/engine/pull/328


JSUYA:
This PR is based on #416.
And
I will reopen as soon as engine PR is ready.

bbrto21:
For this PR, do we need to update tizen version to 6.0 or higher on [tizen_tools](https://github.com/flutter-tizen/tizen_tools)?
([here](https://github.com/flutter-tizen/flutter-tizen/issues/351#issuecomment-1183064479) is the known problem with this.)
This is an important thing and I think we should discuss it in advance before you open the PR.

JSUYA:
> For this PR, do we need to update tizen version to 6.0 or higher on [tizen_tools](https://github.com/flutter-tizen/tizen_tools)? ([here](https://github.com/flutter-tizen/flutter-tizen/issues/351#issuecomment-1183064479) is the known problem with this.) This is an important thing and I think we should discuss it in advance before you open the PR.

This is absolutely necessary as Tizen NUI is supported on version 6.5 or higher.

embedding/csharp/Tizen.Flutter.Embedding/Tizen.Flutter.Embedding.csproj
```diff
@@ -1,12 +1,9 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+﻿  <Project Sdk="Tizen.NET.Sdk/1.1.7">
```
JSUYA, line 1: I need opinion on whether it's okay to modify csproj like this.

swift-kim: I think you should target [multiple TFMs](https://docs.microsoft.com/en-us/nuget/create-packages/multiple-target-frameworks-project-file) in this case, but let's discuss in more detail with @WonyoungChoi tomorrow (offline).

Example: https://stackoverflow.com/questions/42747977/how-do-you-multi-target-a-net-core-class-library-with-csproj

JSUYA: It seems that just modifying the `PackageReference` version is good. Thank you for your reply.
```csproj
```

JSUYA: Umm, I was only saw about the failed build of this PR. For the multiple TFMs you mentioned, I'll look into it too.

swift-kim: I've been testing several combinations and https://github.com/swift-kim/flutter-tizen/blob/multiple-tfms/embedding/csharp/Tizen.Flutter.Embedding/Tizen.Flutter.Embedding.csproj seems to be the best solution for now. I'll discuss further with @WonyoungChoi though.

JSUYA:
@swift-kim PR is ready for review. Please review this.
+)
 This class is written based on ElmFlutterView. After this PR is merged, I want to make an abstraction (maybe FlutterView?) with 
ElmFlutterView class.

JSUYA:
@swift-kim  I need to consider the case when the size is not set. When the patch is ready, I will request a review again.

JSUYA:
> @swift-kim I need to consider the case when the size is not set. When the patch is ready, I will request a review again.
> 
> [JSUYA#1](https://github.com/JSUYA/flutter-tizen/pull/1)



1. If the user sets the size before call `RunEngine()`
 -> Follows the size set in the view. (`Size2D`)

2. If the user sets the size after call `RunEngine()`
 2.1. If there is parent
   -> Sets the size set in the parent view as the initial value of FlutterView.
      (`GetParent().Size2D`)
 2.2. If there is no parent
   -> Sets the size set in the window as the initial value of `FlutterView`.
      (`NUIApplication.GetDefaultWindow().Size`)
 -> Finally, when `Relayout` is called, it follows the determined size.
    (call `FlutterDesktopViewResize()`)

When the `RunEngine()` is called, if the size of the FlutterView is not determined, the Size(parent,window) that can be estimated is calculated and set.
This can be improved when features such as `EngineCache` are added in the future.

JSUYA:
This is NUIFlutterView Sample
https://github.com/JSUYA/NUI_FLUTTER_SAMPLE

```c#
Window window = NUIApplication.GetDefaultWindow();
View rootView = new View()
{
    WidthSpecification = LayoutParamPolicies.MatchParent,
    HeightSpecification = LayoutParamPolicies.MatchParent,
    Layout = new LinearLayout() { LinearOrientation = LinearLayout.Orientation.Vertical }
};

NUIFlutterView flutterView = new NUIFlutterView()
{
    //Relative Layout
    WidthSpecification = LayoutParamPolicies.MatchParent,
    HeightSpecification = LayoutParamPolicies.MatchParent,

    // or Size2D = new Size2D(300, 200)
};
if (flutterView.RunEngine())
{
    GeneratedPluginRegistrant.RegisterPlugins(flutterView.Engine);
    rootView.Add(flutterView);
}
```

embedding/csharp/Tizen.Flutter.Embedding/NUIFlutterView.cs
```diff
@@ -0,0 +1,205 @@
...
+{
+    /// <summary>
+    /// Displays a Flutter screen in a Tizen application.
+    /// </summary>
+    public class NUIFlutterView : ImageView
+    {
+        public NUIFlutterView()
+        {
+            _viewSize = Size2D;
+        }
```
swift-kim, line 24: `_viewSize` will be overwrriten in `RunEngine` so setting the value here seems redundant.

Also all other classes have property and member definitions before the constructor definition so please reorder properly.

embedding/csharp/Tizen.Flutter.Embedding/NUIFlutterView.cs
```diff
@@ -0,0 +1,205 @@
...
+
+        /// <summary>
+        /// Returns the size that can be the default.
+        /// </summary>
+        private Size2D GetDefaultSize()
+        {
+            if (Size2D.Width == 0 || Size2D.Height == 0)
+            {
+                View parent = GetParent() as View;
+                if (!parent)
```
swift-kim, line 131: Can only `View` have a `ImageView` as a child? Are there other `Container` types that can be a parent of `ImageView`?

embedding/csharp/Tizen.Flutter.Embedding/NUIFlutterView.cs
```diff
@@ -0,0 +1,205 @@
...
+                TizenLog.Error("Could not create a Flutter engine.");
+                return false;
+            }
+
+            _viewSize = GetDefaultSize();
+
+            Type baseType = typeof(NativeImageQueue).BaseType.BaseType.BaseType;
+            FieldInfo field = baseType.GetField("swigCPtr", BindingFlags.NonPublic | BindingFlags.Instance);
+            NativeImageQueue nativeImageQueue =
+                new NativeImageQueue((uint)_viewSize.Width, (uint)_viewSize.Height, NativeImageQueue.ColorFormat.RGBA8888);
```
swift-kim, line 82: How about renaming `_viewSize` to just `_size`? Otherwise you need to reformat this line since it exceeds the column length limit (120).

```suggestion
            NativeImageQueue nativeImageQueue = new NativeImageQueue(
                (uint)_viewSize.Width, (uint)_viewSize.Height, NativeImageQueue.ColorFormat.RGBA8888);
```

embedding/csharp/Tizen.Flutter.Embedding/NUIFlutterView.cs
```diff
@@ -0,0 +1,205 @@
...
+                    Window window = NUIApplication.GetDefaultWindow();
+                    return new Size2D(window.Size.Width, window.Size.Height);
+                }
+                return parent.Size2D;
+            }
+            return Size2D;
+        }
+
+        /// <summary>
+        /// Register view event callbacks.
```
swift-kim, line 142: nit:
```suggestion
        /// Registers view event handlers.
```

embedding/csharp/Tizen.Flutter.Embedding/NUIFlutterView.cs
```diff
@@ -0,0 +1,205 @@
...
+        /// </summary>
+        private Size2D GetDefaultSize()
+        {
+            if (Size2D.Width == 0 || Size2D.Height == 0)
+            {
+                View parent = GetParent() as View;
+                if (!parent)
+                {
+                    Window window = NUIApplication.GetDefaultWindow();
+                    return new Size2D(window.Size.Width, window.Size.Height);
```
swift-kim, line 134: May be simplified
```suggestion
                    return window.Size;
```

embedding/csharp/Tizen.Flutter.Embedding/NUIFlutterView.cs
```diff
@@ -0,0 +1,205 @@
...
+                }
+                return parent.Size2D;
+            }
+            return Size2D;
+        }
+
+        /// <summary>
+        /// Register view event callbacks.
+        /// </summary>
+        private void RegisterEvent()
```
swift-kim, line 144:
```suggestion
        private void RegisterEventHandlers()
```

embedding/csharp/Tizen.Flutter.Embedding/NUIFlutterView.cs
```diff
@@ -0,0 +1,205 @@
...
+        {
+            Focusable = true;
+            KeyEvent += (object s, KeyEventArgs e) =>
```
swift-kim, line 147: You might have to unregister the handlers as well.

embedding/csharp/Tizen.Flutter.Embedding/NUIFlutterView.cs
JSUYA: As you said, we don't need this initialization. I removed the constructor.

embedding/csharp/Tizen.Flutter.Embedding/NUIFlutterView.cs
JSUYA: There is Window.Add() case.
In the case of Window.Add(), I used DefaultWindow() because parent cannot be obtained through GetParent().
There are no other types as far as I know.

embedding/csharp/Tizen.Flutter.Embedding/NUIFlutterView.cs
JSUYA: done

embedding/csharp/Tizen.Flutter.Embedding/NUIFlutterView.cs
JSUYA: done

embedding/csharp/Tizen.Flutter.Embedding/NUIFlutterView.cs
JSUYA: done

embedding/csharp/Tizen.Flutter.Embedding/NUIFlutterView.cs
JSUYA: done

embedding/csharp/Tizen.Flutter.Embedding/NUIFlutterView.cs
JSUYA: unregister is unnecessary

embedding/csharp/Tizen.Flutter.Embedding/NUIFlutterView.cs
```diff
@@ -0,0 +1,199 @@
...
+
+            var viewProperties = new FlutterDesktopViewProperties
+            {
+                width = _size.Width,
+                height = _size.Height,
+            };
+
+            View = FlutterDesktopViewCreateFromImageView(
+                ref viewProperties, Engine.Engine, imageViewRef.Handle, nativeImageQueueRef.Handle,
+                NUIApplication.GetDefaultWindow().GetNativeId());
```
swift-kim, line 94: `GetDefaultWindow` seems to be an internal API (it has an `EditorBrowsableState.Never` attribute).

```suggestion
                Window.Instance.GetNativeId());
```

Please also check the line 128.

embedding/csharp/Tizen.Flutter.Embedding/NUIFlutterView.cs
swift-kim: How about the `Layer` class? An instance of it can be obtained through `Window.Instance.GetDefaultLayer()`.

embedding/csharp/Tizen.Flutter.Embedding/NUIFlutterView.cs
JSUYA: wow.. thanks..

swift-kim:
I referred to https://github.com/flutter-tizen/flutter-tizen/pull/418#issuecomment-1228058800 but
```csharp
View rootView = new View()
{
    WidthSpecification = LayoutParamPolicies.MatchParent,
    HeightSpecification = LayoutParamPolicies.MatchParent,
    Layout = new LinearLayout() { LinearOrientation = LinearLayout.Orientation.Vertical }
};
Window.Instance.Add(rootView);

var flutterView = new NUIFlutterView()
{
    WidthSpecification = LayoutParamPolicies.MatchParent,
    HeightSpecification = LayoutParamPolicies.MatchParent,
};
if (flutterView.RunEngine())
{
    GeneratedPluginRegistrant.RegisterPlugins(flutterView.Engine);
    rootView.Add(flutterView);
}
```
shows nothing on the screen. Am I doing something wrong? (`window` is unused in your example.)

embedding/csharp/Tizen.Flutter.Embedding/NUIFlutterView.cs
JSUYA: I added layer case and change to use GetDefaultLayer().

JSUYA:
> I referred to [#418 (comment)](https://github.com/flutter-tizen/flutter-tizen/pull/418#issuecomment-1228058800) but
> 
> ```cs
> View rootView = new View()
> {
>     WidthSpecification = LayoutParamPolicies.MatchParent,
>     HeightSpecification = LayoutParamPolicies.MatchParent,
>     Layout = new LinearLayout() { LinearOrientation = LinearLayout.Orientation.Vertical }
> };
> Window.Instance.Add(rootView);
> 
> var flutterView = new NUIFlutterView()
> {
>     WidthSpecification = LayoutParamPolicies.MatchParent,
>     HeightSpecification = LayoutParamPolicies.MatchParent,
> };
> if (flutterView.RunEngine())
> {
>     GeneratedPluginRegistrant.RegisterPlugins(flutterView.Engine);
>     rootView.Add(flutterView);
> }
> ```
> 
> shows nothing on the screen. Am I doing something wrong? (`window` is unused in your example.)

As far as I know that when I call SetImage(), the ImageView's Size2D is set to the target's Size.
However, this size is set when Relayout() is called.
I modified it to set Size after SetImage. Thank you for your check.

embedding/csharp/Tizen.Flutter.Embedding/NUIFlutterView.cs
```diff
@@ -0,0 +1,211 @@
...
+                    View, type, e.Touch.GetLocalPosition(0).X, e.Touch.GetLocalPosition(0).Y, e.Touch.GetTime(),
+                    e.Touch.GetDeviceId(0));
+
+                _lastTouchEventTime = e.Touch.GetTime();
+                return true;
+            };
+
+            Relayout += (object s, EventArgs e) =>
+            {
+                if (IsRunning && (_size.Width != Size2D.Width || _size.Height != Size2D.Height))
```
swift-kim, line 203: > As far as I know that when I call SetImage(), the ImageView's Size2D is set to the target's Size.
However, this size is set when Relayout() is called.
I modified it to set Size after SetImage.

How about moving this check to the native side (`FlutterTizenView::Resize`) so that you don't have to manage the `_size` value?

JSUYA:
> As far as I know that when I call SetImage(), the ImageView's Size2D is set to the target's Size.
> However, this size is set when Relayout() is called.
> I modified it to set Size after SetImage. Thank you for your check.

After applying this commit there was a problem when placing the flutterview in a relative layout.
I will request a review again after further check.

embedding/csharp/Tizen.Flutter.Embedding/NUIFlutterView.cs
```diff
@@ -0,0 +1,211 @@
...
+        /// </summary>
+        private Size2D GetDefaultSize()
+        {
+            if (Size2D.Width == 0 || Size2D.Height == 0)
+            {
+                Container parent = GetParent();
+                if (parent == null)
+                {
+                    Layer layer = Window.Instance.GetDefaultLayer();
+                    return new Size2D(layer.Viewport.Width, layer.Viewport.Height);
```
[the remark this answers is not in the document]
JSUYA: Size is read only(tizen version 3)
WindowSize can set, get size...(tizen verion 4)
Both are use same internal getter api....

embedding/csharp/Tizen.Flutter.Embedding/NUIFlutterView.cs
JSUYA: I moved the size check code to engine.
https://github.com/flutter-tizen/engine/pull/328/commits/69d461813b19e749581d48b71f7bdc4dfd676f7b

embedding/csharp/Tizen.Flutter.Embedding/NUIFlutterView.cs
```diff
@@ -0,0 +1,211 @@
...
+                }
+
+                if (parent is View)
+                {
+                    View view = parent as View;
```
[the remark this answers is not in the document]
JSUYA: done

embedding/csharp/Tizen.Flutter.Embedding/NUIFlutterView.cs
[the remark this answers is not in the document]
swift-kim: That's not an answer for the question...

embedding/csharp/Tizen.Flutter.Embedding/NUIFlutterView.cs
JSUYA: > I moved the size check code to engine. [flutter-tizen/engine@69d4618](https://github.com/flutter-tizen/engine/commit/69d461813b19e749581d48b71f7bdc4dfd676f7b)

I reverted the engine code and changed the embedding to check the size.
(https://github.com/flutter-tizen/engine/pull/328/commits/7bfd4ea1655e05345d3c60b05d358c71dc2ae01c)

swift-kim:
The following examples generate different results as described below although they are technically equivalent.

ElmSharp:
```csharp
Window window = new Window("ElmSharpApp");
window.BackButtonPressed += (s, e) =>
{
    Exit();
};
window.Show();
```

NUI:
```csharp
Window.Instance.KeyEvent += (s, e) =>
{
    if (e.Key.State == Key.StateType.Down && (e.Key.KeyPressedName == "XF86Back" || e.Key.KeyPressedName == "Escape"))
    {
        Exit();
    }
};
```

In ElmSharp's case, the app is always closed when I press the back button. In NUI's case, the app is closed only when the Flutter view doesn't have a focus. How do you think of this? What would be the correct results for both?

JSUYA:
> In ElmSharp's case, the app is always closed when I press the back button. In NUI's case, the app is closed only when the Flutter view doesn't have a focus. How do you think of this? What would be the correct results for both?

nice catch. NUI is no longer propagated when the `View` consumes the `TouchEvent` even if the `Window` has a `TouchEvent`.
If `View`'s `TouchEvent` returns `false`, the `TouchEvent` is propagated to the event callback of the `View`'s parent.
I think there is something wrong with the Window TouchEvent in the NUI.  I don't know it can be solve in this PR


test code
```c#
ImageView image = new ImageView();
image.Size = new Size(300, 200);
image.SetImage("image path");
rootView.Add(image);
            
image.Focusable = true;
image.KeyEvent += (object s, View.KeyEventArgs e) =>
{
        return true;
};

image.TouchEvent += (object s, View.TouchEventArgs e) =>
{
         FocusManager.Instance.SetCurrentFocusView(image);
         return true;
};

Window.Instance.KeyEvent += (s, e) =>
{
    if (e.Key.State == Key.StateType.Down && (e.Key.KeyPressedName == "XF86Back" || e.Key.KeyPressedName == "Escape"))
     {
           Exit();
     }
};
```

swift-kim (approved):
Regarding the key event issue, I think we need to consider adding support for synchronous keyboard events again. See https://github.com/flutter-tizen/engine/issues/221 and relevant embedder implementations for other platforms (keywords: "delayed event synthesis", "redispatch event"). The work will be later done in a separate PR.

Please merge https://github.com/flutter-tizen/engine/pull/328 and update the engine hash when you're ready. Good job!

JSUYA:
> Regarding the key event issue, I think we need to consider adding support for synchronous keyboard events again. See [flutter-tizen/engine#221](https://github.com/flutter-tizen/engine/issues/221) and relevant embedder implementations for other platforms (keywords: "delayed event synthesis", "redispatch event"). The work will be later done in a separate PR.

I opened new issue page for this.(https://github.com/flutter-tizen/engine/issues/331). I will check it.

> Please merge [flutter-tizen/engine#328](https://github.com/flutter-tizen/engine/pull/328) and update the engine hash when you're ready. Good job!

I added hash update commit.

Thank you for your review.

PR #420, opened 2022-08-12
Inject the TizenApiVersion property when running .NET build

swift-kim:
Will be used by https://github.com/flutter-tizen/flutter-tizen/pull/418 as its implementation depends on Tizen 6.5 APIs.

How to add a file that is not compatible with Tizen 4.0:

```xml
  <ItemGroup Condition="$(TizenApiVersion) &lt; 6.5">
    <PackageReference Include="Tizen.NET" Version="4.0.1.14164" />
    <Compile Remove="NUIFlutterView.cs" />
    <None Include="NUIFlutterView.cs" />
  </ItemGroup>
```

If `<None>` is omitted, the file will not displayed in the project explorer in Visual Studio.


embedding/csharp/Tizen.Flutter.Embedding/Tizen.Flutter.Embedding.csproj
```diff
@@ -5,8 +5,16 @@
     <TargetFramework>netstandard2.0</TargetFramework>
   </PropertyGroup>
 
-  <ItemGroup>
+  <PropertyGroup>
+    <TizenApiLevel Condition="'$(TizenApiLevel)' == ''">4</TizenApiLevel>
```
swift-kim, line 9: Tip: This default value may temporarily be changed to 9 during development (to enable IntelliSense).

lib/tizen_tpk.dart
```diff
@@ -181,6 +181,9 @@ class TizenManifest {
   /// The application type (either "capp" or "dotnet").
   String? get applicationType => _applications.first.getAttribute('type');
 
+  /// The TizenFX API level of the application.
+  String? get dotnetApiLevel => _applications.first.getAttribute('api-version');
```
WonyoungChoi, line 185: Unfortunately,.. `api-version` attribute of `ui-applicatoin` has been deprecated since Tizen 6.5. So currently `tizen-manifest.xml` does not describe `C# API Level` anymore. But we can inter the `API Level` with `api-version` of `manifest` element by using a mapping table. 
refer: https://github.com/Samsung/Tizen.NET/issues/138

WonyoungChoi: How about just passing the platform version to the embedding.csproj? You may use the platform version in conditions.

swift-kim: Fixed. Thank you for the information!

embedding/csharp/Tizen.Flutter.Embedding/Tizen.Flutter.Embedding.csproj
```diff
@@ -5,8 +5,16 @@
     <TargetFramework>netstandard2.0</TargetFramework>
   </PropertyGroup>
 
-  <ItemGroup>
+  <PropertyGroup>
+    <TizenApiVersion Condition="'$(TizenApiVersion)' == ''">4.0</TizenApiVersion>
+  </PropertyGroup>
+
+  <ItemGroup Condition="$(TizenApiVersion) &lt; 6.5">
```
WonyoungChoi, line 12: I think that's enough now. but this is just a reference for later. 
https://docs.microsoft.com/en-us/visualstudio/msbuild/property-functions?view=vs-2019#msbuild-version-comparison-functions

WonyoungChoi (approved):
looks good to me

embedding/csharp/Tizen.Flutter.Embedding/Tizen.Flutter.Embedding.csproj
swift-kim: Oh, the functions look cleaner and more reliable.

PR #422, opened 2022-08-12
Update for embedder API changes

Swanseo0:
- Update `interop.cs` related to changes in `flutter_tizen.h`  (flutter-tizen/engine#314)
- Delete function
  - `FlutterDesktopPluginRegistrarGetNativeWindow`
  - `FlutterDesktopViewGetEvasObject`

- Implement functions
  - `FlutterDesktopPluginRegistrarGetView`
  - `FlutterDesktopViewGetNativeHandle`

Should be merged after [flutter-tizen/engine#314](https://github.com/flutter-tizen/engine/pull/314) is merged


swift-kim:
Did you even test this change? There's an occurrence of `FlutterDesktopViewGetEvasObject` in `embedding/cpp/elm_flutter_view.cc`.

Swanseo0:
> Did you even test this change? There's an occurrence of `FlutterDesktopViewGetEvasObject` in `embedding/cpp/elm_flutter_view.cc`.

Sorry. I missed the change from `FlutterDesktopViewGetEvasObject` to `FlutterDesktopViewGetNativeHandle`.
Can you share the test method or share the relevant documents?
Building a default app with a builded `engine` did not confirm the error.

swift-kim:
Oh, I'm not sure what you mean by "default app" but building a C++ template app or building any plugin example app will definitely fail without your latest force-pushed change. (Make sure to rebase your commit. Your base branch is too old.)

embedding/csharp/Tizen.Flutter.Embedding/Interop/Interop.cs
```diff
@@ -68,7 +68,7 @@ public struct FlutterDesktopEngineProperties
             FlutterDesktopEngine engine);
 
         [DllImport("flutter_tizen.so")]
-        public static extern IntPtr FlutterDesktopPluginRegistrarGetNativeWindow(
+        public static extern FlutterDesktopView FlutterDesktopPluginRegistrarGetView(
```
swift-kim, line 71: You'd just remove this function and `FlutterDesktopViewGetNativeHandle` since they are unused. They can be added later when they are used again.

Swanseo0: There is a [plugin](https://github.com/flutter-tizen/plugins/pull/400) that requires that APIs. 
I think this PR needs to add that APIs.

swift-kim: No. `Interop.cs` is only used within the C# embedding and has nothing to do with plugins.

Swanseo0: Then, as you mentioned, I will remove these APIs from `Interop.cs`.

swift-kim (approved):

PR #423, opened 2022-08-12
Update tizen-manifest.xml of C# app templates

swift-kim:
The `api-version` attribute of the `ui-application` element in `tizen-manifest.xml` is never used and we can safely remove it according to https://github.com/flutter-tizen/flutter-tizen/pull/420#discussion_r944099917.


WonyoungChoi (approved):
LGTM

PR #429, opened 2022-08-22
Support building C# embedding into a nuget package

swift-kim:
- Revert bae818da85b5e02db6308d0b8f32c0bccea2131d and use multiple TFMs instead.
- Running `dotnet build` on the embedding csproj generates a nuget package (which is to be published on [nuget.org](https://nuget.org/)).
- The generated nuget package contains `Tizen.Flutter.Embedding.targets` file which is required for importing Flutter modules from a .NET app. The app's csproj may be configured as follows (assuming that the Flutter module is located in the `../MyFlutterModule` directory).<p>
  ```xml
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>tizen40</TargetFramework>
  </PropertyGroup>

  <PropertyGroup>
    <FlutterModuleDir>../MyFlutterModule</FlutterModuleDir>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Tizen.Flutter.Embedding" Version="1.0.0" />
  </ItemGroup>
  ```
  The `MyFlutterModule` directory contains the following files.<p>
  ```sh
  MyFlutterModule/
  ├── lib
  │   ├── libflutter_engine.so
  │   ├── libflutter_plugins.so
  │   └── libflutter_tizen.so
  ├── res
  │   ├── flutter_assets
  │   │   ├── AssetManifest.json
  │   │   ├── FontManifest.json
  │   │   ├── fonts
  │   │   │   └── MaterialIcons-Regular.otf
  │   │   ├── isolate_snapshot_data
  │   │   ├── kernel_blob.bin
  │   │   ├── NOTICES.Z
  │   │   ├── packages
  │   │   │   └── cupertino_icons
  │   │   │       └── assets
  │   │   │           └── CupertinoIcons.ttf
  │   │   └── vm_snapshot_data
  │   └── icudtl.dat
  └── src
      └── GeneratedPluginRegistrant.cs
  ```


lib/build_targets/package.dart
```diff
@@ -124,10 +124,6 @@ class DotnetTpk extends TizenPackage {
       copyDirectory(pluginsLibDir, libDir);
     }
 
-    final TizenManifest tizenManifest =
-        TizenManifest.parseFromXml(tizenProject.manifestFile);
```
JSUYA, line 128: Is there any way to check the TargetFramework in this dart script?
In #418, I used the way to check the api version in the manifest to use "*_nui.so".

swift-kim: I don't think there is. I think relying on the `api-version` value in the manifest should be enough. Also don't forget to update code in `plugins.dart` (line 92) as well.

embedding/csharp/Tizen.Flutter.Embedding/Tizen.Flutter.Embedding.csproj
```diff
@@ -2,18 +2,30 @@
...
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'tizen40'">
     <PackageReference Include="Tizen.NET" Version="4.0.1.14164" />
   </ItemGroup>
 
-  <ItemGroup Condition="$(TizenApiVersion) &gt;= 6.5">
+  <ItemGroup Condition="'$(TargetFramework)' != 'tizen40'">
     <PackageReference Include="Tizen.NET" Version="9.0.0.16760" />
   </ItemGroup>
```
WonyoungChoi, line 30: I think we don't need to consider the TargetFramework to refer `Tizen.NET` at this time. 
`Tizen.NET` will choose proper `Tizen.NET.APIx` package automatically.

WonyoungChoi: For #418, We may need conditions later to include NUI files with followings..  
- TargetFrameworkIdentifier == 'Tizen' and TargetFrameworkVersion > '9.0'
- TargetFrameworkIdentifier == 'NETStandard' or TargetFrameworkIdentifier == 'NETCoreApp'

swift-kim: Fixed. Thanks! The following lines will be added in the future.

```xml
  <ItemGroup Condition="'$(TargetFramework)' == 'tizen40'">
    <Compile Remove="NUIFlutterView.cs" />
  </ItemGroup>
```

WonyoungChoi (approved):
LGTM

embedding/csharp/Tizen.Flutter.Embedding/Tizen.Flutter.Embedding.csproj
swift-kim: Rebased and added the lines (9f10d2f10261eb95e9465b73e310dc3b9756607e).

PR #431, opened 2022-08-25
Support building a .NET module

swift-kim:
Contributes to https://github.com/flutter-tizen/flutter-tizen/issues/397.

Building a module:

```sh
# Create a module project.
$ flutter-tizen create -t module dotnet_module
$ cd dotnet_module

# Run build (for mobile x86).
$ flutter-tizen build module -pmobile --debug --target-arch x86

# Test without a container app.
$ flutter-tizen run
```

Importing the module from a container app:

```xml
<PropertyGroup>
  <FlutterModuleDir>../dotnet_module/build/tizen/module</FlutterModuleDir>
</PropertyGroup>

<ItemGroup>
  <PackageReference Include="Tizen.Flutter.Embedding" Version="1.0.0" />
</ItemGroup>
```

Note: The `Tizen.Flutter.Embedding` nuget will be published after https://github.com/flutter-tizen/flutter-tizen/pull/429 is merged.

### DIFF
--- a/lib/tizen_plugins.dart
+++ b/lib/tizen_plugins.dart
@@ -469,27 +469,24 @@ using System;
 using System.Runtime.InteropServices;
 using Tizen.Flutter.Embedding;
 
-namespace Runner
+internal class GeneratedPluginRegistrant
 {
-    internal class GeneratedPluginRegistrant
+  {{#cppPlugins}}
+    [DllImport("{{libName}}.so")]
+    public static extern void {{pluginClass}}RegisterWithRegistrar(
+        FlutterDesktopPluginRegistrar registrar);
+
+  {{/cppPlugins}}
+    public static void RegisterPlugins(IPluginRegistry registry)
     {
       {{#cppPlugins}}
-        [DllImport("{{libName}}.so")]
-        public static extern void {{pluginClass}}RegisterWithRegistrar(
-            FlutterDesktopPluginRegistrar registrar);
-
+        {{pluginClass}}RegisterWithRegistrar(
+            registry.GetRegistrarForPlugin("{{pluginClass}}"));
       {{/cppPlugins}}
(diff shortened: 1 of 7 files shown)


templates/module/csharp/.tizen.tmpl/tizen-manifest.xml.tmpl
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<manifest package="{{tizenIdentifier}}" version="1.0.0" api-version="4.0" xmlns="http://tizen.org/ns/packages">
+<manifest package="{{tizenIdentifier}}" version="1.0.0" api-version="6.5" xmlns="http://tizen.org/ns/packages">
```
swift-kim, line 2: We assume that .NET modules are built for Tizen 6.5+ by default (so that an embedder with the `_nui` suffix is used whenever applicable). If you want to use a .NET module on Tizen 6.0 or below, you can change this behavior by manually changing the `api-version` value in `.tizen/tizen-manifest.xml`.

lib/tizen_plugins.dart
```diff
@@ -469,27 +469,24 @@ using System;
 using System.Runtime.InteropServices;
 using Tizen.Flutter.Embedding;
 
-namespace Runner
+internal class GeneratedPluginRegistrant
```
swift-kim, line 472: @WonyoungChoi The `Runner` namespace has been removed from `GeneratedPluginRegistrant` to avoid

```csharp
Runner::GeneratedPluginRegistrant.RegisterPlugins(flutterView.Engine);
```

when importing a module from an app that uses a different namespace. Is it okay to use the global namespace?

WonyoungChoi: Yes, I think it doesn't matter.

PR #433, opened 2022-08-30
Re-enable the attach command

swift-kim:
Add the `attach` command again but with a limitation.

This change is for debugging support of module apps (contributes to https://github.com/flutter-tizen/flutter-tizen/issues/351). The `debug-url` should be manually obtained from the device log (dlog) by the user. I tried to automate the procedure but failed for several reasons:

- `TizenDevice.getLogReader` has not enough information to choose between `TizenDlogReader` and `ForwardingLogReader`.
- `dlogutil -c` and `dlogutil -m` are not supported by non-root devices.
- The device log has incorrect timestamps on TV. Also the `getdate` command doesn't provide any timezone information.

### DIFF
--- a/lib/tizen_device.dart
+++ b/lib/tizen_device.dart
@@ -470,6 +470,9 @@ class TizenDevice extends Device {
 
   @override
   Future<bool> stopApp(TizenTpk app, {String? userIdentifier}) async {
+    if (app == null) {
+      return false;
+    }
     try {
       final List<String> command = usesSecureProtocol
           ? <String>['shell', '0', 'kill', app.applicationId]
--- a/lib/vscode_helper.dart
+++ b/lib/vscode_helper.dart
@@ -68,7 +68,7 @@ void updateLaunchJsonFile(FlutterProject project, Uri observatoryUri) {
     config['cwd'] = project.hasExampleApp
         ? r'${workspaceFolder}/example'
         : r'${workspaceFolder}';
-    config['observatoryUri'] = observatoryUri.toString();
+    config['vmServiceUri'] = observatoryUri.toString();
   }
 
   const JsonEncoder encoder = JsonEncoder.withIndent('    ');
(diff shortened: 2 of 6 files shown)


lib/tizen_device.dart
```diff
@@ -470,6 +470,9 @@ class TizenDevice extends Device {
 
   @override
   Future<bool> stopApp(TizenTpk app, {String? userIdentifier}) async {
+    if (app == null) {
+      return false;
+    }
```
swift-kim, line 475: This check might be removed in the future. `app` may be null when invoked from code without sound null safety (especially when stopping debugging after attach).

lib/vscode_helper.dart
```diff
@@ -68,7 +68,7 @@ void updateLaunchJsonFile(FlutterProject project, Uri observatoryUri) {
     config['cwd'] = project.hasExampleApp
         ? r'${workspaceFolder}/example'
         : r'${workspaceFolder}';
-    config['observatoryUri'] = observatoryUri.toString();
+    config['vmServiceUri'] = observatoryUri.toString();
```
swift-kim, line 71: `"observatoryUri"` is deprecated.

bbrto21 (approved):

lib/vscode_helper.dart
bbrto21: This term is still used in the document(such as `The observatory URL`, `--observatory-uri`). is it okay?

swift-kim: What I meant was that the config property had been deprecated by the VS Code extension.

image

https://github.com/flutter/devtools/issues/1642

The **Dart extension** uses the term "VM service", but the term "observatory" still also seems valid across the **Flutter SDK** (the code and documentation). I'm not sure about DevTools' case but anyway I'm trying to be consistent with the Flutter SDK as much as possible.

PR #434, opened 2022-08-31
Publish C# embedding as a nuget on nuget.org

swift-kim:
The workflow is only run manually on demand (`workflow_dispatch`).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,13 +7,13 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ github.repository_owner == 'flutter-tizen' }}
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-dotnet@v1
+      - uses: actions/checkout@v3
+      - uses: actions/setup-dotnet@v2
         with:
           dotnet-version: '6.0.x'
       - name: Build and push
+        working-directory: embedding/csharp/Tizen.Flutter.Embedding
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,21 @@
...
+
+on: [workflow_dispatch]
+
+jobs:
+  nuget:
+    runs-on: ubuntu-latest
+    if: ${{ github.repository_owner == 'flutter-tizen' }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-dotnet@v1
```
WonyoungChoi, line 11: It is recommended to use the latest version.

```suggestion
      - uses: actions/checkout@v3
      - uses: actions/setup-dotnet@v2
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,21 @@
...
+        with:
+          dotnet-version: '6.0.x'
+      - name: Build and push
+        run: |
+          cd embedding/csharp/Tizen.Flutter.Embedding
```
WonyoungChoi, line 16: How about using `working-directory` property?
```suggestion
      - name: Build and push
        working-directory: embedding/csharp/Tizen.Flutter.Embedding
        run: |
```

.github/workflows/release.yml
swift-kim: Fixed. Thanks!

WonyoungChoi (approved):
LGTM :+1:

PR #439, opened 2022-09-15
Copy a correct embedder binary during .NET module build

swift-kim:
Fixes a bug reported by @bbrto21 where the tool incorrectly copies `libflutter_tizen_mobile.so` instead of `libflutter_tizen_mobile_nui.so` during .NET module build (targeting Tizen 6.5).

The relevant code has accidentally been removed during rebase (due to a conflict between https://github.com/flutter-tizen/flutter-tizen/pull/431 and https://github.com/flutter-tizen/flutter-tizen/pull/418).


JSUYA (approved):

bbrto21 (approved):
Thank you for quick fix :)

Issue #456, opened 2022-11-03
Wrong assembly referenced when building an NUI app


swift-kim:
When the `-o` (output) option is provided to the `dotnet build` command,

https://github.com/flutter-tizen/flutter-tizen/blob/940c2f3018f9ff3c26157a8224c44c54b16e0193/lib/build_targets/package.dart#L140-L147

embedding assemblies for each target framework (`netstandard2.0`, `tizen40`, `tizen90`) are copied to the specified directory in an **arbitrary** order and thus unexpectedly replace each other. This results in a wrong assembly being referenced at build time or runtime as follows.

- Without the `-o` option: The assemblies are copied to their corresponding target framework directories. (SUCCESS)
  ```
  MSBuild version 17.3.0+92e077650 for .NET
    Determining projects to restore...
    All projects are up-to-date for restore.
    Tizen.Flutter.Embedding -> /home/swift/Git/flutter-tizen/embedding/csharp/Tizen.Flutter.Embedding/bin/Debug/netstandard2.0/Tizen.Flutter.Embedding.dll
    Tizen.Flutter.Embedding -> /home/swift/Git/flutter-tizen/embedding/csharp/Tizen.Flutter.Embedding/bin/Debug/tizen90/Tizen.Flutter.Embedding.dll
    Tizen.Flutter.Embedding -> /home/swift/Git/flutter-tizen/embedding/csharp/Tizen.Flutter.Embedding/bin/Debug/tizen40/Tizen.Flutter.Embedding.dll
  ...
  ```

- With the `-o` option: The assemblies are copied to the same output directory. (FAIL)
  ```
  Exception: Failed to build .NET application:
  MSBuild version 17.3.0+92e077650 for .NET
    Determining projects to restore...
    All projects are up-to-date for restore.
    Tizen.Flutter.Embedding -> /home/swift/Git/dummy/nui_module/build/tizen/tpk/Tizen.Flutter.Embedding.dll
    Tizen.Flutter.Embedding -> /home/swift/Git/dummy/nui_module/build/tizen/tpk/Tizen.Flutter.Embedding.dll
    Tizen.Flutter.Embedding -> /home/swift/Git/dummy/nui_module/build/tizen/tpk/Tizen.Flutter.Embedding.dll
  /home/swift/Git/dummy/nui_module/tizen/App.cs(10,9): error CS0246: The type or namespace name 'NUIFlutterView' could not be found (are you missing a using directive or an assembly
[...]

PR #457, opened 2022-11-03
Do not use the output option when running .NET CLI build

swift-kim:
Fixes https://github.com/flutter-tizen/flutter-tizen/issues/456.

Do not use the `-o` (output) option but use the default .NET CLI build directory. The path to the build directory is parsed from the CLI output since the target framework name (e.g. `tizen40`) is unknown to the flutter-tizen tool. The build artifacts (TPK file and tpkroot) are copied to the output directory (`build/tizen/tpk`) after build.


WonyoungChoi (approved):
LGTM 👍

PR #460, opened 2022-11-08
Introduce the flutter_tizen package

swift-kim:
Preparation for https://github.com/flutter-tizen/plugins/issues/399#issuecomment-1298139243.

As of now, the only public API of the package is `isTizen` which returns true if the calling process is running on Tizen.

@bbrto21 How do you think of this approach? As you can see, the implementation is quite tricky (all contents of `flutter-tizen/packages/*` are copied to `flutter/packages` on every run since `dart pub` searches for SDK packages only in that path and we can't override it). Another solution is to publish the `flutter_tizen` package on pub.dev. In that way the package can be properly version controlled using semantic versioning but the user has to specify a package version in their pubspec:

```patch
@@ -10,8 +10,7 @@ dependencies:

   flutter:
     sdk: flutter
-  flutter_tizen:
-    sdk: flutter
+  flutter_tizen: ^1.0.0

 environment:
   sdk: ">=2.13.0 <3.0.0"
```

p.s. The macOS test is failing for unknown reason (the `TizenPub` constructor not being called).


swift-kim:
Beware that the package name conflicts with the [tool package](https://github.com/flutter-tizen/flutter-tizen/blob/master/pubspec.yaml) name.

bbrto21:
That's a cool solution. but Is it possible to use a path rather than copying a package?
If the tool can automatically insert a dependency with path when creating a project, I expect we don't have to worry about the version and the implementation will be simple on the tool side.

```
flutter_tizen: 
  path: ~/FLUTER_TIZEN_PATH/package/flutter_tizen
```

swift-kim:
@bbrto21 `pubspec.yaml` files are subject to version control and should not be arbitrarily modified by the tool.

bbrto21:
@swift-kim  Then I agree that this draft is the best approach at this point. Could you please proceed this PR?

.github/workflows/analyze.yml
```diff
@@ -18,9 +16,9 @@ jobs:
       - name: Install pub dependencies
         run: flutter/bin/flutter pub get
       - name: Verify formatting
-        run: flutter/bin/dart format --output=none --set-exit-if-changed lib test
+        run: flutter/bin/dart format --output=none --set-exit-if-changed lib test packages
       - name: Analyze project source
-        run: flutter/bin/dart analyze --fatal-infos lib
+        run: flutter/bin/dart analyze --fatal-infos lib test
```
swift-kim, line 21: It's very tricky to run analysis on the packages directory so I decided to verify only formatting.

bbrto21 (approved):
LGTM

PR #470, opened 2022-12-06
Migrate test code to null safety

swift-kim:


JSUYA (approved):
`// @dart = 2.8` remains in `./bin/flutter_tizen.dart`. don't you remove it?

swift-kim:
@JSUYA flutter_tools' `executable.dart` has not fully been migrated to sound null safety and the following error is generated when generating a tool snapshot.

```sh
Compiling flutter-tizen...
Info: Compiling with sound null safety
../../.pub-cache/hosted/pub.dartlang.org/dwds-15.0.0/lib/dwds.dart:5:1: Error: A library can't opt out of null safety by default, when using sound null safety.
// @dart = 2.9
^^^^^^^^^^^^^^
../../.pub-cache/hosted/pub.dartlang.org/dwds-15.0.0/lib/src/connections/debug_connection.dart:5:1: Error: A library can't opt out of null safety by default, when using sound null safety.
// @dart = 2.9
^^^^^^^^^^^^^^
...
```

There are a few ways to work around the issue but I don't think it's very necessary at this moment.

PR #476, opened 2023-01-05
Add more tests to tizen_device_test.dart

swift-kim:
Add basic tests for `TizenDevice.installApp()`, `TizenDevicePortForwarder`.

### DIFF
--- a/test/general/tizen_device_test.dart
+++ b/test/general/tizen_device_test.dart
@@ -69,7 +69,7 @@ void main() {
         stdout: <String>[
           'cpu_arch:armv7',
           'secure_protocol:disabled',
-          'platform_version:4.0',
+          'platform_version:5.5',
         ].join('\n'),
       ),
       FakeCommand(
@@ -111,7 +111,7 @@ void main() {
   });
 
   testWithoutContext(
-      'TizenDevice.installApp uninstalls and reinstalls TPK if installation fails',
+      'TizenDevice.installApp warns if the device API version is lower than the package API version',
       () async {
     final TizenDevice device = _createTizenDevice(
       processManager: processManager,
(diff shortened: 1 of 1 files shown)


test/general/tizen_device_test.dart
```diff
@@ -207,6 +280,61 @@ void main() {
...
+  testWithoutContext(
+      'TizenDevicePortForwarder.forwardedPorts can list forwarded ports', () {
+    final TizenDevicePortForwarder forwarder = TizenDevicePortForwarder(
+      device: _createTizenDevice(
+        processManager: processManager,
+        fileSystem: fileSystem,
+      ),
+      logger: logger,
+    );
+    processManager.addCommands(<FakeCommand>[
```
bbrto21, line 293: If you use `addCommand` instead of `addCommand`s, this part can be shorter.

bbrto21 (approved):
Generally LGTM.

test/general/tizen_device_test.dart
swift-kim: Oh, you're right. I just forgot.

test/general/tizen_device_test.dart
```diff
@@ -106,6 +110,75 @@ void main() {
...
+        command: _sdbCommand(<String>['uninstall', 'TestPackage']),
+        stdout: '''
+__return_cb req_id[1] pkg_type[tpk] pkgid[TestPackage] key[install_percent] val[100]
+__return_cb req_id[1] pkg_type[tpk] pkgid[TestPackage] key[end] val[ok]
+''',
+      ),
+      FakeCommand(command: _sdbCommand(<String>['install', 'app.tpk'])),
+    ]);
+
+    expect(await device.installApp(tpk), isTrue);
```
JSUYA, line 176: How about setting the value of `apiVersion` in `_FakeTizenManifest`?
You can increase the coverage of the `installApp()` function by one more line.

[tizen_device.dart]
```
     260                 :          2 :     final Version? platformVersion = Version.parse(_platformVersion);
     261                 :          3 :     final Version? apiVersion = Version.parse(app.manifest.apiVersion);
     262                 :            :     if (platformVersion != null &&
     263                 :            :         apiVersion != null &&
     264                 :          0 :         apiVersion > platformVersion) {   // <---- here
     265                 :          0 :       _logger.printWarning(
```

JSUYA (approved):

test/general/tizen_device_test.dart
swift-kim: Thanks. I added a test for the version comparison.

PR #488, opened 2023-03-30
Make dart_api_dl.h symbols available to app and plugin code

swift-kim:
Currently it's not very easy to access the Dart VM API from plugin native code because the relevant header and source files must be included in each plugin's source tree. This patch automatically adds `dart_api_dl.c` to the `NativeEmbedding` build target (so it's statically linked with the plugin code) and adds the `dart_api_dl.h` location to the include path during build, making the APIs available to all app and plugin native code:

```cpp
#include <dart_api_dl.h>

intptr_t ImagePickerTizenPluginInitializeDartApi(void *data) {
  return Dart_InitializeApiDL(data);
}
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,7 +12,9 @@ jobs:
         with:
           channel: stable
       - name: Install credentials
-        run: echo '${{ secrets.PUB_CREDENTIALS }}' > $PUB_CACHE/credentials.json
+        run: |
+          mkdir -p $PUB_CACHE
+          echo '${{ secrets.PUB_CREDENTIALS }}' > $PUB_CACHE/credentials.json
       - name: Dry run on pull request
         if: ${{ github.event_name == 'pull_request' }}
         run: |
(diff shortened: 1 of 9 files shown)


bbrto21:
I try to use this PR at existing project , but I got linking error, but including headers is fine.

swift-kim:
@bbrto21 `dart_api_dl.c` should still exist in your plugin source tree and compiled along with the plugin.

swift-kim:
I will fix this error soon.

```sh
error: invalid argument '-std=c++17' not allowed with 'C'
make[1]: *** [/home/swift/tizen-studio/tools/ide/resources/native/Build/makefile.mk:96: Debug/src/dart_api_dl.o] Error 1
```

bbrto21 (approved):

bbrto21:
Suddenly, I came up with a problem that multiple plugins can try init.

swift-kim:
@bbrto21 I've confirmed that there's no problem with building an application that depends on multiple plugins, where each plugin contains the `dart_api_dl.c` source file, because `dart_api_dl.h` symbols are not exported by the plugins. Anyway, I'll still try the "implicit" build solution to simplify the things.

bbrto21 (approved):

PR #495, opened 2023-05-04
PlatformView: Add setOffset to PostFrameCallback

Swanseo0:
`PlatformView` is being created when `_viewController.setSize()` is called. By the way, we didn't know `offset` because the platform was created with only size.
This `offset` is obtained by calling RenderTizenView.localToGlobal(Offset.zero) after _viewController.setSize().


packages/flutter_tizen/lib/src/rendering/platform_view.dart
```diff
@@ -118,6 +118,7 @@ class RenderTizenView extends PlatformViewRenderBox {
     do {
       targetSize = size;
       _currentTextureSize = await _viewController.setSize(targetSize);
+      _viewController.setOffset(localToGlobal(Offset.zero));
```
swift-kim, line 121: What's the relationship between the resize and offset operations? Why should they performed together? Why didn't you add a post frame callback just like `RenderAndroidView`?

swift-kim (commented):
The package version should be increased in order to release this change.

Also a colon should not be preceded by a space, as I mentioned before. Please don't forget again.

https://github.com/flutter-tizen/embedder/pull/25#discussion_r1111453071
https://github.com/flutter-tizen/embedder/pull/25#discussion_r1102407948

packages/flutter_tizen/lib/src/rendering/platform_view.dart
Swanseo0: > What's the relationship between the resize and offset operations? Why should they performed together?

`create(size, offset)` takes size and offset as arguments. However, when `_viewController.setSize(targetSize)` is called, only size is used to create a platform view(`create(size)`).  After the platform view is created, the platform view is notified of the offset.
> Why didn't you add a post frame callback just like RenderAndroidView?

The use of postframe callback has been confirmed to work as I intended. I'll apply the code. Thank you.

swift-kim: > `create(size, offset)` takes size and offset as arguments. However, when `_viewController.setSize(targetSize)` is called, only size is used to create a platform view(`create(size)`). After the platform view is created, the platform view is notified of the offset.

This doesn't answer my question. I was asking why you called `setOffset` inside the method (`_sizePlatformView`) whose role is to "resize" the platform view but not to set its offset.

swift-kim (approved):

packages/flutter_tizen/lib/src/rendering/platform_view.dart
```diff
@@ -127,6 +129,17 @@ class RenderTizenView extends PlatformViewRenderBox {
     markNeedsPaint();
   }
 
+  void _setOffset() {
+    SchedulerBinding.instance.addPostFrameCallback((_) async {
+      if (!_isDisposed) {
+        if (attached) {
+          await _viewController.setOffset(localToGlobal(Offset.zero));
+        }
+        _setOffset();
```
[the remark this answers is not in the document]
bbrto21: I can't remember, but I think it's because it's basically based on the platform view that was in our web view implementation. so at the time, it seems like no effort was made to copy the comments from the original Flutter implementation.

bbrto21 (approved):

PR #497, opened 2023-05-18
Upgrade to Flutter 3.10

swift-kim:
* Update the framework and engine hashes to target Flutter 3.10.1.
* Update the tool source code. See the commit history for details.


swift-kim:
The `commands/test_test.dart` fails on Windows due to [this recent upstream change](https://github.com/flutter/flutter/commit/06952ba2549ec009fd82d24f571fdb9c263e2326). The `.toFilePath()` method appends `\` to the directory path, but later `globals.fs.path.absolute` appends `/` to the path again, resulting in a final path that looks like `/\integration_test` which is not recognized as a directory path.

This looks like a bug but I will not file an issue for now.

Issue #283, opened 2021-11-29
Can plugin specify required version of flutter-tizen (sdk)?


wiertel:
Is it possible for a plugin (not sure it would be useful for apps) to declare a version of flutter-tizen/flutter-tizen sdk/engine that it requires/is compatible with?

After a change in engine (flutter-tizen/engine#209) the webview_flutter_tizen has applied it (flutter-tizen/plugins#278). And then, when I tried to build an application with updated plugin I got a C++ compilation error. That was fixed by updating flutter-tizen. The build break in application is justified and understandable. But it would be a lot nicer if there was short error, possibly with helpful information.
I guess that for showing something like this before trying to build the app would need the plugin to declare dependency/require/compatible version of flutter-tizen.

swift-kim:
Yeah, it's a known issue, and unfortunately there's no way to define a compatible flutter-tizen version inside a plugin (only Dart and Flutter versions can be specified), because the flutter-tizen tool and engine are even not "versioned" linearly. (We currently manage [released engine](https://github.com/flutter-tizen/engine/releases) versions only using hash values.) The current best practice is not to make such a breaking API change between the embedder and plugins, and if it is made, notify the users of the plugins to use the latest available flutter-tizen tool.

Does anybody have a good idea to deal with this issue?

wiertel:
It makes sense to manage the changes as you described for now. Maybe it is not even worth implementing anything else for that, as breaking changes would not be often. Or maybe could be coupled with main flutter sdk updates, so existing mechanism could be used.
By existing mechanism I mean [SDK constraints](https://dart.dev/tools/pub/pubspec#sdk-constraints).
Ideas:
1. The fact that engine is not versioned is a major obstacle. Maybe build number could be used? Those are sequential and one can compare any 2 of them.
2. Contacting dart maintainers to discuss extending those SDK constraints with tizen engine information. They could either apply it, or suggest other way. We cannot simply add something to  `environment` as the structure is verified and it is processed by `pub get` etc - so outside of our reach.
3. Introduce a "virtual package" - either really virtual that the flutter-tizen would detect and react specially, or a real package with only purpose is to check current engine and fail if there is something wrong. The failure should result in nice error message or else there would be no benefit of the package.implement 
[...]

PR #501, opened 2023-06-07
Set FLUTTER_BUILD_DIR environment variable when building plugins

swift-kim:
Set the `FLUTTER_BUILD_DIR` environment variable during plugin and app build to allow some files to be downloaded to the location when building FlutterFire plugins.

cc @daeyeon @hs0225

### DIFF
--- a/lib/build_targets/package.dart
+++ b/lib/build_targets/package.dart
@@ -397,7 +397,7 @@ class NativeTpk extends TizenPackage {
       },
       sign: securityProfile,
       environment: <String, String>{
-        'FLUTTER_BUILD_DIR': environment.buildDir.path.toPosixPath(),
+        'FLUTTER_BUILD_DIR': environment.buildDir.path.toPosixPath(''),
       },
     );
     if (result.exitCode != 0) {
--- a/lib/build_targets/package.dart
+++ b/lib/build_targets/package.dart
@@ -396,6 +396,9 @@ class NativeTpk extends TizenPackage {
         'targets': <String>['b1'],
       },
       sign: securityProfile,
+      environment: <String, String>{
+        'FLUTTER_BUILD_DIR': environment.buildDir.path.toPosixPath(),
+      },
     );
     if (result.exitCode != 0) {
       throwToolExit('Failed to build native application:\n$result');
--- a/lib/build_targets/plugins.dart
+++ b/lib/build_targets/plugins.dart
@@ -142,6 +142,9 @@ class NativePlugins extends Target {
           ],
         ],
         rootstrap: rootstrap.id,
+        environment: <String, String>{
+          'FLUTTER_BUILD_DIR': environment.buildDir.path.toPosixPath(),
+        },
       );
       if (result.exitCode != 0) {
         throwToolExit('Failed to build ${plugin.name} plugin:\n$result');
--- a/lib/build_targets/plugins.dart
+++ b/lib/build_targets/plugins.dart
@@ -143,7 +143,7 @@ class NativePlugins extends Target {
         ],
         rootstrap: rootstrap.id,
         environment: <String, String>{
-          'FLUTTER_BUILD_DIR': environment.buildDir.path.toPosixPath(),
+          'FLUTTER_BUILD_DIR': environment.buildDir.path.toPosixPath(''),
         },
       );
       if (result.exitCode != 0) {
--- a/lib/tizen_sdk.dart
+++ b/lib/tizen_sdk.dart
@@ -150,6 +150,7 @@ class TizenSdk {
     String? output,
     Map<String, Object> package = const <String, Object>{},
     String? sign,
+    Map<String, String> environment = const <String, String>{},
   }) {
     String stringify(Object argument) {
       if (argument is String) {
@@ -185,6 +186,7 @@ class TizenSdk {
       environment: <String, String>{
         'PATH': getPathVariable(),
         'USER_CPP_OPTS': '-std=c++17',
+        ...environment,
       },
     );
   }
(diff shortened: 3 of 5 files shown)


daeyeon:
Thanks a lot. I will check if it works well for our case.

daeyeon:
Observed two issues below.

1) `FLUTTER_BUILD_DIR` is not passed as an environment variable when building apps with `project_def.prop`.

```make
# project_def.prop
USER_INC_DIRS = inc src ${FLUTTER_BUILD_DIR}/.firebaseSDK
```

```console
"/home/daeyeon/tizen-studio/tools/smart-build-interface/../llvm-10/bin/clang++" -c "src/cloud_functions_plugin.cc" -o "Debug/src/cloud_functions_plugin.o" -DTIZEN_DEPRECATION -DDEPRECATION_WARNING -DMOBILE_PROFILE
-DFLUTTER_PLUGIN_IMPL -DTIZEN -D__TIZEN__

-Iinc -Isrc -I/.firebaseSDK/inc  <-

-I"pch" -g3 -D_DEBUG -O0 -Wall -c
```

2) The value of `FLUTTER_BUILD_DIR` begins and ends with double quotes. This may look a bit unusual and cause unexpected behaviour when concatenating strings, so I tested it by removing the quotes them as shown below.

```bash
echo ${HOME}/.firebaseSDK
# /home/daeyeon/.firebaseSDK

echo ${FLUTTER_BUILD_DIR}
# "/home/daeyeon/code/sec.plugins/packages/firebase_tests/.dart_tool/flutter_build/aff5146beb803517879986d57df211ea"

echo ${FLUTTER_BUILD_DIR}/.firebaseSDK
# "/home/daeyeon/code/sec.plugins/packages/firebase_tests/.dart_tool/flutter_build/aff5146beb803517879986d57df211ea"/.firebaseSDK

echo ${FLUTTER_BUILD_DIR:1:-1}/.firebaseSDK
# /home/daeyeon/code/sec.plugins/packages/firebase_tests/.dart_tool/flutter_build/aff5146beb803517879986d57df211ea/.firebaseSDK
```

swift-kim:
@daeyeon

> `FLUTTER_BUILD_DIR` is not passed as an environment variable when building apps with `project_def.prop`.

I think you need to use valid Makefile notation. Please refer to https://www.gnu.org/software/make/manual/make.html and try replacing the curly braces with parentheses.

> The value of `FLUTTER_BUILD_DIR` begins and ends with double quotes.

The directory path may contain spaces and you need to use double quotes to safely escape them. I'll remove the quotes from the variable value but you need to use them properly in your shell script.

daeyeon:
There was a mistake in my previous observation. I don't know why, but both `${FLUTTER_BUILD_DIR}` and `$(FLUTTER_BUILD_DIR)` work. Sorry for confusing. I managed to build an app with `FLUTTER_BUILD_DIR`, but having trouble with a directory path that contains spaces. The following is observed when building an app in `/home/daeyeon/code/sec.plugins/packages/my test` on my machine.

```make
#project_def.prop
USER_INC_DIRS = inc src dep/include $(FLUTTER_BUILD_DIR)
```

```console
# build logs
...
 -I"/home/daeyeon/code/sec.plugins/packages/my -Itest/.dart_tool/flutter_build/8d81ef7ad3993e1f5b384e4f9fd8b91b" 
...
```
I attempted builds with enclosing the above `$(FLUTTER_BUILD_DIR)` in quotes, double quotes, and several forms of encapsulation, but the result is unchanged. Additionally, setting the include path or anything else through `USER_CPPFLAGS_MISC` doesn't seem to work at all. :thinking:

daeyeon:
> I think you need to use valid Makefile notation. Please refer to https://www.gnu.org/software/make/manual/make.html and try replacing the curly braces with parentheses.

6.1 Basics of Variable References
To substitute a variable’s value, write a dollar sign followed by the name of the variable in parentheses or braces: **either ‘$(foo)’ or ‘${foo}’ is a valid reference to the variable foo.** This special significance of ‘$’ is why you must write ‘$$’ to have the effect of a single dollar sign in a file name or recipe.

swift-kim:
@daeyeon 

> I attempted builds with enclosing the above `$(FLUTTER_BUILD_DIR)` in quotes, double quotes, and several forms of encapsulation, but the result is unchanged.

Tizen Studio supports [escaping spaces in `project_def.prop` files](https://github.sec.samsung.net/RS-TizenStudio/home/issues/201) using the `\` character as of Tizen Studio 4.5.

> Additionally, setting the include path or anything else through `USER_CPPFLAGS_MISC` doesn't seem to work at all.

As far as I know the `USER_CPPFLAGS_MISC` property in `project_def.prop` is [no longer supported](https://github.com/flutter-tizen/flutter-tizen/pull/493) by Tizen Studio. You could use the [`USER_CPP_OPTS` environment variable](https://github.com/flutter-tizen/flutter-tizen/blob/07c795b9a0634e983ce1fc68d4ccae88c24fc6da/lib/tizen_sdk.dart#L187) but make sure not to omit `-std=c++17` (which is mandatory for plugin build) when overriding the value.

daeyeon:
Confirmed that my application was built successfully with https://github.com/flutter-tizen/flutter-tizen/pull/501/commits/b9b6279c846ff2139e71624be409126e7c5f351b. @swift-kim Can you let us know the minimum required SDK version for this feature once it's available? I believe the version needs to be specified in `pubspec.yaml`.

swift-kim:
@daeyeon The flutter-tizen tool has no formal versioning policy, and actually [there is no way](https://github.com/flutter-tizen/flutter-tizen/issues/283) to specify its version in `pubspec.yaml`. The feature will be available (to those using the latest version of the tool) once this PR is merged.

daeyeon (approved):
this LGTM, but it would be good to get other folks chiming in as well.

daeyeon (approved):

PR #508, opened 2023-08-25
Fix flutter_tools.stamp version check

swift-kim:
The `flutter` command is being called redundantly due to https://github.com/flutter/flutter/commit/2835c3f1161afcd93894324475c72eb66f67a05e.

### DIFF
--- a/bin/internal/shared.sh
+++ b/bin/internal/shared.sh
@@ -49,7 +49,7 @@ function update_flutter() {
   if [[ "$version" != "$(git rev-parse HEAD)" ]]; then
     git reset --hard
     git clean -xdf
-    git fetch "$FLUTTER_REPO" "$version"
+    git fetch --tags "$FLUTTER_REPO" "$version"
     git checkout FETCH_HEAD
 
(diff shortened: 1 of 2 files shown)


bin/internal/shared.sh
```diff
@@ -66,8 +66,9 @@ function update_flutter() {
   unset GIT_WORK_TREE
 
   # Invalidate the flutter cache.
+  local compilekey="$version:"
```
JSUYA, line 69: `.stamp` does not contain `:`. Will the stamp information also change? Or is below commands no longer needed?

JSUYA (approved):

bin/internal/shared.sh
swift-kim: No. The format has already changed as you can see in the above linked commit. We just didn't notice because there was no visible difference (due to `> /dev/null`).
```sh
$ cat flutter/bin/cache/flutter_tools.stamp
e1e47221e86272429674bec4f1bd36acc4fc7b77:
```

PR #509, opened 2023-09-04
Pass API_VERSION environment variable to the builder

swift-kim:
A feature requested by our SRC-N members. @xiaowei-guan

- Sets the `API_VERSION` environment variable (which is in the `x.y` format) when building native plugins and apps.
- Searches the `lib/{BUILD_ARCH}/{API_VERSION}` directory for platform-version specific user libraries.


xiaowei-guan (approved):
LGTM

PR #514, opened 2023-09-26
Add C# multi app support

swift-kim:
- Add the `multi-app/csharp` template.
- Remove unnecessary code (infolink elements) and file (icon) from existing templates.
- Support `flutter-tizen build` and `flutter-tizen run`.
- Remove support for building a .NET TPK with the default certificate (to reduce complexity).


swift-kim:
Need to update the wiki when merged.

haesik:
Does this PR include all the necessary functions?

lib/build_targets/package.dart
```diff
@@ -129,7 +129,44 @@ class DotnetTpk extends TizenPackage {
...
+          .printStatus('The $securityProfile profile is used for signing.');
+    } else {
+      throwToolExit(
+        'No certificate profile found. You can create one using the Certificate Manager.\n'
+        'https://github.com/flutter-tizen/flutter-tizen/blob/master/doc/install-tizen-sdk.md#create-a-tizen-certificate',
+      );
+    }
+
+    File locateTpk(String stdout) {
+      // "Runner -> /path/to/project/tizen/bin/Debug/tizen40/com.example.app-1.0.0.tpk"
```
JSUYA, line 157: The path depending on the app type and build type (debug, release). I think this is an unnecessary comment.
or `/path/to/project/tizen/bin/{App and build type}/tizen40/com.example.app-1.0.0.tpk"` ?

test/commands/create_test.dart
```diff
@@ -107,21 +107,6 @@ void main() {
     expect(projectDir.childFile('tizen/tizen-manifest.xml'), exists);
   }, overrides: <Type, Generator>{});
```
JSUYA, line 109: How about add create test?
```dart
testUsingContext('Can create a C# multi app project', () async {
    final TizenCreateCommand command = TizenCreateCommand();
    final CommandRunner<void> runner = createTestCommandRunner(command);
    await runner.run(<String>[
      'create',
      '--no-pub',
      '--platforms=tizen',
      '--app-type=multi',
      '--tizen-language=csharp',
      projectDir.path,
    ]);

    final File mainDart = projectDir.childFile('lib/main.dart');
    expect(mainDart.readAsStringSync(), contains('Tizen Multi App Demo'));

    final String rawPubspec =
        projectDir.childFile('pubspec.yaml').readAsStringSync();
    final Pubspec pubspec = Pubspec.parse(rawPubspec);
    expect(pubspec.dependencies, contains('tizen_app_control'));
    expect(pubspec.dependencies, contains('messageport_tizen'));

    expect(projectDir.childDirectory('tizen/ui').listSync(), isNotEmpty);
    expect(projectDir.childDirectory('tizen/service').listSync(), isNotEmpty);

    final ProcessResult result = await Process.run(
      'git',
      <String>['diff', '--name-only'],
      workingDirectory: Cache.flutterRoot,
    );
    expect(result.exitCode, 0);
    expect(result.stdout, isEmpty);
  }, overrides: <Type, Generator>{});
```

JSUYA (approved):

lib/build_targets/package.dart
swift-kim: This just represents the output pattern for the regular expression in the next line. Here's an example of the MSBuild output.

```sh
$ dotnet build tizen/service/
MSBuild version 17.3.2+561848881 for .NET
  Determining projects to restore...
  All projects are up-to-date for restore.
  Tizen.Flutter.Embedding -> /home/swift/Git/flutter-tizen/embedding/csharp/Tizen.Flutter.Embedding/bin/Debug/tizen40/Tizen.Flutter.Embedding.dll
  RunnerService -> /home/swift/Git/dummy/csharp_multi_app/tizen/service/bin/Debug/tizen40/RunnerService.dll
  Configuration :
  Platform :
  TargetFramework :
  TizenTpkFiles : tizen-manifest.xml
  RunnerService is signed with Default Certificates!
  RunnerService -> /home/swift/Git/dummy/csharp_multi_app/tizen/service/bin/Debug/tizen40/com.example.csharp_multi_app-1.0.0.tpk

Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:02.14
```

We only parse the part `RunnerService -> ...` from this output. `RunnerService`, `Debug`, `tizen40`, `com.example.csharp_multi_app-1.0.0` are all subject to change and do not matter.

test/commands/create_test.dart
swift-kim: I thought that would be a redundant test case. Compared to the C++ multi app creation test, the only different part is `--tizen-language=csharp`, which doesn't affect the code coverage. If we're to add a test, it would be better to add a test case to `package_test.dart` instead, but it's too much work for me.

xuelian-bai (approved):

Issue #510, opened 2023-09-08
defaultTargetPlatform returns TargetPlatform.android in release build


JSUYA:
When building with `--release` option, `defaultTargetPlatform` returns `TargetPlatform.android` rather than `TargetPlatform.linux`.
(In debug build, it returns `TargetPlatform.linux`.)
```
import 'package:flutter/foundation.dart';
...
print(defaultTargetPlatform.toString());
...
```

Swanseo0:
IMO, this seems to be a problem because `android_arm` is being targeted when creating `TizenAotElf` in `release mode`.

According to [build_info.dart](https://github.com/flutter/flutter/blob/b90a13f3a1383229daf7a3f9e696e966391dba59/packages/flutter_tools/lib/src/build_info.dart#L528), Unfortunately, `linux_arm` is not supported. So we're using `android_arm` to build `app.so`.

swift-kim:
This problem may have been introduced by [`@pragma("vm:platform-const")`](https://github.com/dart-lang/sdk/commit/2c0484c720cdbcfa9463f89629f9c10c775a9ca9) which can be found in several places in [`dart-sdk/lib/io/platform.dart`](https://github.com/dart-lang/sdk/blob/main/sdk/lib/io/platform.dart). I'm not sure how I can rebuild the framework snapshot without the annotations. I tried manually building `vm_isolate_snapshot.bin` and `isolate_snapshot.bin` from source (the `generate_snapshot_bins` ninja target) and replacing existing files in `flutter/bin/cache/artifacts/engine/linux-x64` but it didn't work.

The corresponding change in the upstream tool code can be found in https://github.com/flutter/flutter/commit/840d7dd68a7af5fb660e80ca5a923e0adccb26b8. A possible (but disruptive) workaround for this issue is to force replace the `targetOS` argument in [`common.dart`](https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/build_system/targets/common.dart) when building the kernel snapshot as follows:

```diff
@@ -249,7 +263,7 @@ class KernelSnapshot extends Target {
       dartDefines: decodeDartDefines(environment.defines, kDartDefines),
       packageConfig: packageConfig,
       buildDir: environment.buildDir,
-      targetOS: targetOS,
+      targetOS: 'linux',
       checkDartPluginRegistry: environment.generateDartPluginRegistry,
     );
     if (output == null || output.errorCount != 0) {
```
[...]

PR #515, opened 2023-10-11
Override targetOS when building KernelSnapshot

swift-kim:
A temporary solution for https://github.com/flutter-tizen/flutter-tizen/issues/510.

Passes `linux` (originally `android`) as the `targetOS` argument to `KernelCompiler.compile`.


JSUYA (approved):
LGTM

PR #516, opened 2023-10-11
Make Dart plugin registrants work on background isolates

swift-kim:
Fixes https://github.com/flutter-tizen/plugins/issues/123.

Implements https://github.com/flutter/flutter/issues/98591 for Tizen.

Previously, an app that depends on any _Dart plugin_ (a platform package of a federated plugin that has any Dart implementation) had to explicitly call the `register` method of each plugin to use the plugin in a _background isolate_ as follows.

```dart
compute((_) async {
  PathProviderPlugin.register();

  print(await getApplicationDocumentsDirectory());
}, null);
```

Now this platform-specific call is no longer necessary, and instead the app can simply call Flutter's `DartPluginRegistrant.ensureInitialized` to initialize all Dart plugins that it depends on.

```dart
compute((_) async {
  DartPluginRegistrant.ensureInitialized();

  print(await getApplicationDocumentsDirectory());
}, null);
```

For details, see also:
- `FindAndInvokeDartPluginRegistrant` in [runtime/dart_plugin_registrant.cc](https://github.com/flutter/engine/blob/main/runtime/dart_plugin_registrant.cc)
- `dartPluginRegistrantLibrary` in [dart_plugin_registrant.dart](https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/dart_plugin_registrant.dart)
- `DartPluginRegistrant` in [dart:ui/natives.dart](https://github.com/flutter/engine/blob/main/lib/ui/natives.dart)

Known issue: This still does not work after a Hot restart.

Note that if the app wants to use _platform channels_ in a background isolate, it should additionally call `BackgroundIsolateBinaryMessenger.ensureInitialized` to initialize the engine's binary messenger as follows.

```dart
compute((RootIsolateToken rootToken) async {
  BackgroundIsolateBinaryMessenger.ensureInitialized(rootToken);
  // DartPluginRegistrant.ensureInitialized(); // called implicitly by the above

  ...
}, RootIsolateToken.instance!);
```

### DIFF
--- a/lib/tizen_builder.dart
+++ b/lib/tizen_builder.dart
@@ -19,9 +19,12 @@ import 'package:flutter_tools/src/build_system/build_system.dart';
 import 'package:flutter_tools/src/cache.dart';
 import 'package:flutter_tools/src/commands/assemble.dart';
 import 'package:flutter_tools/src/commands/build_ios_framework.dart';
+import 'package:flutter_tools/src/compile.dart';
+import 'package:flutter_tools/src/dart/package_map.dart';
 import 'package:flutter_tools/src/globals.dart' as globals;
 import 'package:flutter_tools/src/linux/build_linux.dart';
 import 'package:flutter_tools/src/project.dart';
+import 'package:package_config/package_config.dart';
 
 import 'build_targets/package.dart';
 import 'tizen_build_info.dart';
@@ -38,6 +41,7 @@ TizenBuilder? get tizenBuilder => context.get<TizenBuilder>();
 /// - [BuildIOSFrameworkCommand._produceAppFramework] in `build_ios_framework.dart` (build target)
 /// - [AssembleCommand.runCommand] in `assemble.dart` (performance measurement)
 /// - [buildLinux] in `build_linux.dart` (code size)
+/// - [KernelCompiler.compile] in `compile.dart` (Dart plugin registrant)
 class TizenBuilder {
   TizenBuilder();
 
@@ -65,6 +69,23 @@ class TizenBuilder {
     final BuildInfo buildInfo = tizenBuildInfo.buildInfo;
     final String targetPlatform = getNameForTargetPlatform(
         getTargetPlatformForArch(tizenBuildInfo.targetArch));
+
+    // The generated main contains the Dart plugin registrant.
+    final File dartPluginRegistrant = globals.fs.file(targetFile);
+    final PackageConfig packageConfig = await loadPackageConfigWithLogging(
+      project.packageConfigFile,
+      logger: globals.logger,
+    );
+    String? dartPluginRegistrantUri;
+    if (dartPluginRegistrant.existsSync()) {
+      final Uri dartPluginRegistrantFileUri = dartPluginRegistrant.uri;
+      dartPluginRegistrantUri =
+          packageConfig.toPackageUri(dartPluginRegistrantFileUri)?.toString() ??
+              dartPluginRegistrantFileUri.toString();
+    }
+    // See the engine's FindAndInvokeDartPluginRegistrant() in `dart_plugin_registrant.cc`.
+    buildInfo.dartDefines
+        .add('flutter.dart_plugin_registrant=$dartPluginRegistrantUri');
 
     final Environment environment = Environment(
       projectDir: project.directory,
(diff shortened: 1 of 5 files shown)


swift-kim:
Regarding the Hot restart issue:

The best solution would be to set `dartPluginRegistrantLibrary` to a constant pointing to the current file when generating the `generated_main.dart`. Otherwise it would be necessary to override `BuildSystem.buildIncremental` (called by `ResidentRunner.runSourceGenerators`) to append a `flutter.dart_plugin_registrant` value to `dartDefines`.

The full call stack is as follows when a Hot restart is requested and `DefaultResidentCompiler.recompile` is called.

```dart
DefaultResidentCompiler.recompile (\home\swift\Git\flutter-tizen\flutter\packages\flutter_tools\lib\src\compile.dart:677)
DevFS.update (\home\swift\Git\flutter-tizen\flutter\packages\flutter_tools\lib\src\devfs.dart:617)
FlutterDevice.updateDevFS (\home\swift\Git\flutter-tizen\flutter\packages\flutter_tools\lib\src\resident_runner.dart:581)
HotRunner._updateDevFS (\home\swift\Git\flutter-tizen\flutter\packages\flutter_tools\lib\src\run_hot.dart:493)
HotRunner._restartFromSources (\home\swift\Git\flutter-tizen\flutter\packages\flutter_tools\lib\src\run_hot.dart:581)
HotRunner._fullRestartHelper (\home\swift\Git\flutter-tizen\flutter\packages\flutter_tools\lib\src\run_hot.dart:794)
HotRunner.restart (\home\swift\Git\flutter-tizen\flutter\packages\flutter_tools\lib\src\run_hot.dart:734)
TerminalHandler._commonTerminalInputHandler (\home\swift\Git\flutter-tizen\flutter\packages\flutter_tools\lib\src\resident_runner.dart:1791)
TerminalHandler.processTerminalInput (\home\swift\Git\flutter-tizen\flutter\packages\flutter_tools\lib\src\resident_runner.dart:1831)
```

We could even add a hook to overwrite the Flutter plugin registrant (`.dart_tool/flutter_build/dart_plugin_registrant.dart`) before `DefaultResidentCompiler.recompile` is called, by overriding either
- `BuildSystem.buildIncremental`
- `HotRunnerConfig.setupHotRestart` (called by `HotRunner._fullRestartHelper`)
- `FlutterDevice.updateDevFS` (called by `HotRunner._updateDevFS`) (is it possible to override `FlutterDevice`?)

so that we don't have to manage our own `generated_main.dart`.

The former seems pretty straightforward. The latter will take some time to implement and test.

JSUYA:
> edit) It's not possible to set `dartPluginRegistrantLibrary` in the generated main. `flutter/src/dart_plugin_registrant.dart` is hard-coded in `FindAndInvokeDartPluginRegistrant`.

I have a question. (I don't 100% understand this problem yet...)
Is it possible to overwrite the key value of `flutter.dart_plugin_registrant` using --dart-define?

swift-kim:
@JSUYA Yes, the value added to `buildInfo.dartDefines` in `TizenBuilder` is passed to the `KernelCompiler` and that's how this PR works. However it looks like the value is only persistent during the first build but not during subsequent recompiles (hot restart). After reading your comment, I think it would be better to override the `FlutterCommand.getBuildInfo` method (used by `flutter-tizen run`) to append the value so that the new `buildInfo` takes effect globally (including recompiles).

swift-kim:
I tested https://github.com/swift-kim/flutter-tizen/commit/4ae555a23348a45b0935f41de6f67d3fb401c518 but it didn't work (`dartPluginRegistrantLibrary` is not recognized after a hot restart even though the Dart define is set). I'm not sure why.

JSUYA:
> I tested [swift-kim@4ae555a](https://github.com/swift-kim/flutter-tizen/commit/4ae555a23348a45b0935f41de6f67d3fb401c518) but it didn't work (`dartPluginRegistrantLibrary` is not recognized after a hot restart even though the Dart define is set). I'm not sure why.

When I tested this code with re-start, it worked fine. What should be the problem occurred?
```
compute((_) async {
      DartPluginRegistrant.ensureInitialized();
      print(await getTemporaryDirectory());
    }, null);
```

swift-kim:
@JSUYA The test code is not very different from mine. I tested again and it works now (maybe because I ran `flutter-tizen clean`?)

JSUYA (approved):

PR #522, opened 2023-11-16
Fix OverflowException in StandardMessageHelper

Vovanella95:
0xffff = 65535
Int16.MaxValue = 32767

So all platform messages, that has size between 32767 and 65532 will raise exception `System.OverflowException: Value was either too large or too small for an Int16`

This PR fix this problem

### DIFF
--- a/embedding/csharp/Tizen.Flutter.Embedding/Channels/StandardMessageHelper.cs
+++ b/embedding/csharp/Tizen.Flutter.Embedding/Channels/StandardMessageHelper.cs
@@ -68,7 +68,7 @@ namespace Tizen.Flutter.Embedding
             else
             {
                 stream.WriteByte(255);
-                WriteBytes(stream, BitConverter.GetBytes(Convert.ToInt32(size)));
+                WriteBytes(stream, BitConverter.GetBytes(Convert.ToUInt32(size)));
             }
         }
 
--- a/embedding/csharp/Tizen.Flutter.Embedding/Channels/StandardMessageHelper.cs
+++ b/embedding/csharp/Tizen.Flutter.Embedding/Channels/StandardMessageHelper.cs
@@ -60,7 +60,7 @@ namespace Tizen.Flutter.Embedding
             {
                 stream.WriteByte(Convert.ToByte(size));
             }
-            else if (size <= 0xffff)
+            else if (size <= Int16.MaxValue)
             {
                 stream.WriteByte(254);
                 WriteBytes(stream, BitConverter.GetBytes(Convert.ToInt16(size)));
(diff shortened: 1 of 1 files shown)


embedding/csharp/Tizen.Flutter.Embedding/Channels/StandardMessageHelper.cs
lenchis001, line 1: Could you please replace hardcoded value for Byte as well?

Vovanella95: Done

embedding/csharp/Tizen.Flutter.Embedding/Channels/StandardMessageHelper.cs
```diff
@@ -60,7 +60,7 @@ public static void WriteSize(Stream stream, int size)
             {
                 stream.WriteByte(Convert.ToByte(size));
             }
-            else if (size <= 0xffff)
+            else if (size <= Int16.MaxValue)
```
swift-kim, line 63: The size itself seems correct and consistent with the other language ([C++](https://github.com/flutter/engine/blob/0c57a50810e8f2e0f278722db9e28fae67adb4f5/shell/platform/common/client_wrapper/standard_codec.cc#L246-L251)) implementations. I think the `Convert.ToInt16` part should be changed to `Convert.ToUInt16` to match the C++ implementation (same for `Convert.ToInt32`).

Vovanella95: Will check how it works and fix it soon!

Vovanella95: Hi @swift-kim, should I add `Convert.ToUInt32(size)` also?

swift-kim: Yes.

Vovanella95: @swift-kim updated! Could you please merge my fix? (If no other remarks)

swift-kim: Hmm. What I meant by "the two lines" were actually lines 66 and 88, but let me fix this later.

swift-kim (approved):
Thanks!

embedding/csharp/Tizen.Flutter.Embedding/Channels/StandardMessageHelper.cs
Vovanella95: @swift-kim sorry for misunderstanding

PR #534, opened 2024-03-12
Support Tizen 8.0 build

JSUYA:
Starting with Tizen 8.0, the unified "tizen" profile is used.

### DIFF
--- a/lib/tizen_sdk.dart
+++ b/lib/tizen_sdk.dart
@@ -260,11 +260,6 @@ class TizenSdk {
     }
     apiVersion ??= '5.5';
 
-    if (apiVersion == '8.0') {
-      // Note: From tizen-8.0, profile has been integrated into the `tizen`.
-      profile = 'tizen';
-    }
-
     double versionToDouble(String versionString) {
       final double? version = double.tryParse(versionString);
       if (version == null) {
@@ -285,6 +280,11 @@ class TizenSdk {
         apiVersion = '6.0';
       }
       type = 'device64';
+    }
+
+    if (apiVersion == '8.0') {
+      // Note: Starting with Tizen 8.0, the unified "tizen" profile is used.
+      profile = 'tizen';
     }
 
     Rootstrap getRootstrap(String profile, String apiVersion, String type) {
--- a/lib/tizen_sdk.dart
+++ b/lib/tizen_sdk.dart
@@ -260,6 +260,11 @@ class TizenSdk {
     }
     apiVersion ??= '5.5';
 
+    if (apiVersion == '8.0') {
+      // Note: From tizen-8.0, profile has been integrated into the `tizen`.
+      profile = 'tizen';
+    }
+
     double versionToDouble(String versionString) {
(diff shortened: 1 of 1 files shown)


lib/tizen_sdk.dart
```diff
@@ -260,6 +260,11 @@ class TizenSdk {
     }
     apiVersion ??= '5.5';
 
+    if (apiVersion == '8.0') {
+      // Note: From tizen-8.0, profile has been integrated into the `tizen`.
```
swift-kim, line 264: nit: For clarity
```suggestion
      // Note: Starting with Tizen 8.0, the unified "tizen" profile is used.
```

lib/tizen_sdk.dart
```diff
@@ -271,7 +276,7 @@ class TizenSdk {
     String type = arch == 'x86' ? 'emulator' : 'device';
     if (arch == 'arm64') {
       // The arm64 build is only supported by iot-headed-6.0+ rootstraps.
-      if (profile != 'iot-headed') {
+      if (profile != 'tizen' && profile != 'iot-headed') {
         _logger.printError(
             'The arm64 build is not supported by the $profile profile.');
         profile = 'iot-headed';
```
swift-kim, line 282: You may move lines 263-266 after line 288 so that the profile is not overwritten at line 282, or just remove the check (lines 279-283) for simplicity.

lib/tizen_sdk.dart
```diff
@@ -260,6 +260,11 @@ class TizenSdk {
     }
     apiVersion ??= '5.5';
 
+    if (apiVersion == '8.0') {
+      // Note: From tizen-8.0, profile has been integrated into the `tizen`.
+      profile = 'tizen';
```
swift-kim, line 265: Did you check if the instruction at lines 311-312 is still valid?

JSUYA: You mean this command?
`/home/junsu/tizen-studio/package-manager/package-manager-cli.bin install TIZEN-8.0-NativeAppDevelopment-CLI`
it works well in my environment. is there any thing that i have to check more?

lib/tizen_sdk.dart
JSUYA: done

lib/tizen_sdk.dart
JSUYA: I moved if-condition to later. This change is no longer needed.

lib/tizen_sdk.dart
swift-kim: No problem if it works. Thanks.

lib/tizen_sdk.dart
swift-kim: You may still need the check. If the user specified the profile as `tizen`, the tool will emit the error. Yes, it's quite complicated and unreadable. This is why I recommended removing the part.

JSUYA: I modified it. thanks

swift-kim (approved):

Issue #538, opened 2024-04-15
Support tizen 8.0 emulator


JSUYA:
A crash occurs when running the flutter-tizen app on the emulator of the common unified profile of Tizen 8.0.

https://developer.tizen.org/blog/announcing-tizen-studio-5.5-release?langredirect=1
```
  - Added common unified profile called Tizen and deprecated mobile and wearable profiles from Tizen 8.0 version onwards.
```

log
```
E/ConsoleMessage(11271): FlutterApplication.cs: Run(131) > Unhandled exception: System.DllNotFoundException: Unable to load shared library 'flutter_tizen.so' or one of its dependencies. In order to help diagnose loading problems, consider setting the LD_DEBUG environment variable: libflutter_tizen.so.so: cannot open shared object file: No such file or directory
E/ConsoleMessage(11271):    at Tizen.Flutter.Embedding.Interop.FlutterDesktopEngineCreate(FlutterDesktopEngineProperties& engine_properties)
E/ConsoleMessage(11271):    at Tizen.Flutter.Embedding.FlutterEngine..ctor(String assetsPath, String icuDataPath, String aotLibraryPath, String dartEntrypoint, List`1 dartEntrypointArgs) in /home/junsu/dev/os/f-project/flutter-tizen/embedding/csharp/Tizen.Flutter.Embedding/FlutterEngine.cs:line 50
E/ConsoleMessage(11271):    at Tizen.Flutter.Embedding.FlutterEngine..ctor(String dartEntrypoint, List`1 dartEntrypointArgs) in /home/junsu/dev/os/f-project/flutter-tizen/embedding/csharp/Tizen.Flutter.Embedding/FlutterEngine.cs:line 22
E/ConsoleMessage(11271):    at Tizen.Flutter.Embedding.FlutterApplication.OnCreate() in /home/junsu/dev/os/f-project/flutter-tizen/embedding/csharp/Tizen.Flutter.Embedding/FlutterApplication.cs:line 152
E/ConsoleMessage(11271):    at Runner.App.OnCreate() in /home/junsu/dev/os/f-project/dev/test_cs_14/tizen/App.cs:line 9
E/ConsoleMessage(11271):    at Tizen.Applications.CoreBackend.UICoreBackend.OnCreateNative(IntPtr data) in Tizen.Applications.UI.dll:token 0x6000011+0x44

```

JSUYA:
https://github.com/flutter-tizen/flutter-tizen/pull/537

PR #539, opened 2024-04-17
Use 8.0 embedder in Tizen 8.0 application

JSUYA:
In Tizen 8.0, cbhm library has been deprecated.
So we need to use an embedder that does not depend on the cbhm library.

related issue: https://github.com/flutter-tizen/flutter-tizen/issues/538


JSUYA:
This patch should be first. https://github.com/flutter-tizen/embedder/pull/63

Issue #545, opened 2024-05-20
Migrate to Flutter 3.22


JSUYA:
https://medium.com/flutter/io24-5e211f708a37

https://github.com/flutter/flutter/commit/5dcb86f68f239346676ceb1ed1ea385bd215fba1
https://github.com/flutter/engine/commit/f6344b75dcf861d8bf1f1322780b8811f982e31a

PR #551, opened 2024-05-28
Upgrade to Flutter 3.22

JSUYA:
Targeting Flutter 3.22.0.
https://github.com/flutter-tizen/flutter-tizen/issues/545


JSUYA:
rebased

bin/internal/flutter.version
```diff
@@ -1 +1 @@
-ba393198430278b6595976de84fe170f553cc728
+5dcb86f68f239346676ceb1ed1ea385bd215fba1
```
swift-kim, line 1: Can we update the framework and engine to the latest ([3.22.2](https://github.com/flutter/flutter/blob/master/docs/releases/Hotfixes-to-the-Stable-Channel.md#flutter-322-changes))?

JSUYA: ok. I will do this after updating the engine and embedder to the latest.

JSUYA: done

swift-kim (approved):

PR #559, opened 2024-06-12
Refactor if chains into switch statements

swift-kim:
https://github.com/flutter/flutter/pull/144905

### DIFF
--- a/lib/forwarding_log_reader.dart
+++ b/lib/forwarding_log_reader.dart
@@ -67,16 +67,14 @@ class ForwardingLogReader extends DeviceLogReader {
       return message;
     }
     final String prefix = match.group(1)!;
-    TerminalColor color;
-    if (prefix == '[I]') {
-      color = TerminalColor.cyan;
-    } else if (prefix == '[W]') {
-      color = TerminalColor.yellow;
-    } else if (prefix == '[E]') {
-      color = TerminalColor.red;
-    } else if (prefix == '[F]') {
-      color = TerminalColor.magenta;
-    } else {
+    final TerminalColor? color = switch (prefix) {
+      '[I]' => TerminalColor.cyan,
+      '[W]' => TerminalColor.yellow,
+      '[E]' => TerminalColor.red,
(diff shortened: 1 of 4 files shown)


lib/forwarding_log_reader.dart
```diff
@@ -67,16 +67,14 @@ class ForwardingLogReader extends DeviceLogReader {
...
+    final TerminalColor? color = switch (prefix) {
+      '[I]' => TerminalColor.cyan,
+      '[W]' => TerminalColor.yellow,
+      '[E]' => TerminalColor.red,
+      '[F]' => TerminalColor.magenta,
+      _ => null,
+    };
+    if (color == null) {
       return message;
     }
```
JSUYA, line 79: Refactoring, which changes an if chain to a switch, targets enum values.
https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#avoid-using-if-chains-or--or--with-enum-values
Are these changes just for looks better? (readability?)

JSUYA (commented):
https://github.com/flutter-tizen/flutter-tizen/blob/master/lib/tizen_build_info.dart#L23
It seems that getArchForTargetPlatform() should also be modified.

lib/forwarding_log_reader.dart
swift-kim: > getArchForTargetPlatform() should also be modified.

It was a little tricky due to the static analysis rules [`non_exhaustive_switch_statement`](https://dart.dev/tools/diagnostic-messages?utm_source=dartdev&utm_medium=redir&utm_id=diagcode&utm_content=non_exhaustive_switch_statement#non_exhaustive_switch_statement) and [`no_default_cases`](https://dart.dev/tools/linter-rules/no_default_cases) but now I found a workaround.

> targets enum values.

The set of PRs I linked above apply the rule to non-enum values (strings) as well. I was just following their practice. I think it's because it makes the code shorter.

JSUYA (approved):

PR #569, opened 2024-10-31
Update Tizen.NET API version to tizen80

JSUYA:
flutter-tizen supports Tizen 6.0 or higher. Therefore, update the Tizen.NET API version.
(The [TargetFramework](https://github.com/Samsung/TizenFX) for Tizen 6.0 is tizen80.)


packages/flutter_tizen/example/tizen/Runner.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>tizen40</TargetFramework>
+    <TargetFramework>tizen80</TargetFramework>
```
swift-kim, line 5: Changing the TFM may have a side effect. Is there any API you want to use but is not available in API4?

packages/flutter_tizen/example/tizen/Runner.csproj
```diff
@@ -16,4 +16,4 @@
     </TizenTpkUserIncludeFiles>
   </ItemGroup>
 
-</Project>
+</Project>
```
swift-kim, line 19: Unexpectedly removing EOF newlines.

embedding/csharp/Tizen.Flutter.Embedding/Tizen.Flutter.Embedding.csproj
swift-kim, line 1: The nuget version at line 12 should be increased (to 1.2.0).

packages/flutter_tizen/example/tizen/Runner.csproj
JSUYA: As far as I know, there were no APIs that were previously used that are scheduled to be deprecated. And since NUIFlutterView.cs is only used in tizen90(tizen6.5), there were no side effects yet.
When I created a new project and tested it, it worked without any problems. Are there any APIs you're concerned about?

packages/flutter_tizen/example/tizen/Runner.csproj
JSUYA: done

embedding/csharp/Tizen.Flutter.Embedding/Tizen.Flutter.Embedding.csproj
JSUYA: Thank you for your review. I updated PR.

embedding/csharp/Tizen.Flutter.Embedding/Tizen.Flutter.Embedding.csproj
```diff
@@ -28,7 +28,7 @@
     <PackageReference Include="Tizen.NET" Version="9.0.0.16760" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'tizen40'">
+  <ItemGroup Condition="'$(TargetFramework)' == 'tizen40' Or '$(TargetFramework)' == 'tizen80'">
```
seungsoo47, line 31: I have a question.
Why didn't you check 'tizen90'?
I asked because the TargetFrameworks above also included tizen90.

JSUYA: As mentioned in [the comment above](https://github.com/flutter-tizen/flutter-tizen/pull/569#discussion_r1823768904), NUIFlutterView.cs is a feature for Tizen 6.5. Tizen 6.5 is using tizen90. https://github.com/Samsung/TizenFX

seungsoo47: I understood. Thanks for explaining.

packages/flutter_tizen/example/tizen/Runner.csproj
swift-kim: I can't say what exact effect there will be but .NET's target framework resolution is quite complicated. The project may not build if the app depends on a library with an incompatible TFM or the embedding is added to an NUI project as a module. It shouldn't be a problem in general, but I just wanted to warn you about the risk.

swift-kim (approved):

PR #574, opened 2024-12-06
Add prefer_nuget_cache metadata to tizen-manifest.xml

JSUYA:
Add metadata to enable caching for Tizen assembly/library cache(TLC, TAC).
It is known that the TAC on TV devices caches Tizen.Flutter.Embedding by default.
In the tizen profile (RPI), so files (TLC) are applied starting with Tizen 10.

### DIFF
--- a/templates/module/csharp/.tizen.tmpl/tizen-manifest.xml.tmpl
+++ b/templates/module/csharp/.tizen.tmpl/tizen-manifest.xml.tmpl
@@ -5,7 +5,7 @@
         <label>{{projectName}}</label>
         <icon>ic_launcher.png</icon>
         <metadata key="http://tizen.org/metadata/prefer_dotnet_aot" value="true"/>
-        <metadata key="http://tizen.org/metadata/prefer_nuget_cache" value="true" />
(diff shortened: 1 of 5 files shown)


templates/module/csharp/.tizen.tmpl/tizen-manifest.xml.tmpl
```diff
@@ -5,6 +5,7 @@
         <label>{{projectName}}</label>
         <icon>ic_launcher.png</icon>
         <metadata key="http://tizen.org/metadata/prefer_dotnet_aot" value="true"/>
+        <metadata key="http://tizen.org/metadata/prefer_nuget_cache" value="true" />
```
swift-kim, line 8: No space for consistency.
```suggestion
        <metadata key="http://tizen.org/metadata/prefer_nuget_cache" value="true"/>
```

JSUYA: done thanks!

seungsoo47 (approved):

PR #583, opened 2025-01-13
Add enable_impeller metadata key

JSUYA:
(Currently, impeller is experimental step in flutter-tizen.)
The flutter-tizen tool supports --enable-impeller flag.
However, when running an already installed application with app select,
impeller cannot be applied. So, add a metadata key to turn impeller on and off in tizen-manifest.xml.

```
<metadata key="http://tizen.org/metadata/flutter_tizen/enable_impeller" value="true" />
```

### DIFF
--- a/embedding/cpp/flutter_engine.cc
+++ b/embedding/cpp/flutter_engine.cc
@@ -5,6 +5,8 @@
 #include "include/flutter_engine.h"
 
 #include <app.h>
+#include <app_info.h>
+#include <app_manager.h>
 
 #include <cerrno>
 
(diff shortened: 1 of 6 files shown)


embedding/csharp/Tizen.Flutter.Embedding/FlutterApplication.cs
```diff
@@ -150,6 +150,11 @@ protected override void OnCreate()
                 throw new Exception("External output is not supported by FlutterRendererType::kEGL type renderer.");
             }
 
+            if (RendererType == FlutterRendererType.EvasGL && Engine.IsImpellerEnabled)
+            {
+                throw new Exception("Impeller is is not supported by FlutterRendererType::kEvasGL type renderer.");
```
seungsoo47, line 155:
```suggestion
                throw new Exception("Impeller is not supported by FlutterRendererType::kEvasGL type renderer.");
```

seungsoo47 (approved):

JSUYA:
I applied the same logic to cpp application.

embedding/cpp/flutter_engine.cc
```diff
@@ -79,6 +79,7 @@ std::vector<std::string> ParseEngineArgs() {
     auto engine_args_it =
         std::find(engine_args.begin(), engine_args.end(), "--enable-impeller");
     if (engine_args_it == engine_args.end() && it->second == "true") {
+      *is_impeller_enabled = true;
```
seungsoo47, line 82: In case of (engine_args_it != engine_args.end() && it->second == "true"), *is_impeller_enabled is also true, right?

PR #597, opened 2025-03-18
Generate a tizen flutter plugins extra info file

seungsoo47:
This file is generated to help VD division collect Flutter Apps from the App Store. 
It is generated during app build and copied to TPK.

### DIFF
--- a/lib/build_targets/package.dart
+++ b/lib/build_targets/package.dart
@@ -112,8 +112,10 @@ class DotnetTpk extends TizenPackage {
     // provided at runtime, so the file name must be a constant.
     embedder.copySync(libDir.childFile('libflutter_tizen.so').path);
     icuData.copySync(resDir.childFile(icuData.basename).path);
-    appDepsJson.copySync(
-        tizenProject.hostAppRoot.childFile(appDepsJson.basename).path);
+    appDepsJson.copySync(resDir
+        .childDirectory('flutter_assets')
+        .childFile(appDepsJson.basename)
+        .path);
 
     if (buildMode.isPrecompiled) {
       final File aotSnapshot = environment.buildDir.childFile('app.so');
@@ -292,8 +294,10 @@ class NativeTpk extends TizenPackage {
     engineBinary.copySync(libDir.childFile(engineBinary.basename).path);
     embedder.copySync(libDir.childFile(embedder.basename).path);
     icuData.copySync(resDir.childFile(icuData.basename).path);
-    appDepsJson.copySync(
-        tizenProject.hostAppRoot.childFile(appDepsJson.basename).path);
+    appDepsJson.copySync(resDir
+        .childDirectory('flutter_assets')
+        .childFile(appDepsJson.basename)
+        .path);
 
     if (buildMode.isPrecompiled) {
       final File aotSnapshot = environment.buildDir.childFile('app.so');
--- a/lib/build_targets/package.dart
+++ b/lib/build_targets/package.dart
@@ -105,16 +105,15 @@ class DotnetTpk extends TizenPackage {
     final File engineBinary = engineDir.childFile('libflutter_engine.so');
     final File embedder = embedderDir.childFile('libflutter_tizen_$profile.so');
     final File icuData = engineDir.childFile('icudtl.dat');
-    final File pluginsExtraInfo =
-        project.directory.childFile('.flutter-tizen-plugins-extra-info');
+    final File appDepsJson = project.directory.childFile('app.deps.json');
 
     engineBinary.copySync(libDir.childFile(engineBinary.basename).path);
     // The embedder so name is statically defined in C# code and cannot be
     // provided at runtime, so the file name must be a constant.
     embedder.copySync(libDir.childFile('libflutter_tizen.so').path);
     icuData.copySync(resDir.childFile(icuData.basename).path);
-    pluginsExtraInfo
-        .copySync(ephemeralDir.childFile(pluginsExtraInfo.basename).path);
+    appDepsJson.copySync(
+        tizenProject.hostAppRoot.childFile(appDepsJson.basename).path);
 
     if (buildMode.isPrecompiled) {
       final File aotSnapshot = environment.buildDir.childFile('app.so');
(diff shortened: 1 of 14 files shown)


lib/build_targets/package.dart
```diff
@@ -102,16 +102,19 @@ class DotnetTpk extends TizenPackage {
         getEngineArtifactsDirectory(buildInfo.targetArch, buildMode);
     final Directory embedderDir =
         getEmbedderArtifactsDirectory(apiVersion, buildInfo.targetArch);
-
     final File engineBinary = engineDir.childFile('libflutter_engine.so');
     final File embedder = embedderDir.childFile('libflutter_tizen_$profile.so');
     final File icuData = engineDir.childFile('icudtl.dat');
+    final File pluginsExtraInfo =
+        project.directory.childFile('.flutter-tizen-plugins-extra-info');
```
JSUYA, line 109: I think it would be better if the .flutter-tizen-plugins-extra-info file is located in the app root rather than flutter/ephemeral. (In the case of C#, Runner.deps.json is located in the app root)
and I think `.json` extension could kept.
so the file name could be app.deps.json, flutter-tizen.deps.json or something better. If you have a better idea, please let me know.

lib/tizen_plugins.dart
```diff
@@ -629,3 +633,35 @@ Future<void> renderTemplateToFile(
...
+  result['info'] =
+      'This is a generated file; do not edit or check into version control.';
+  result['plugins'] = pluginInfo;
+  result['date_created'] = globals.systemClock.now().toString();
+  result['dart_version'] = globals.flutterVersion.dartSdkVersion;
+  result['flutter_version'] = globals.flutterVersion.frameworkVersion;
+  result['engine_version'] =
+      globals.cache.getStampFor(kTizenEngineStampName) ?? '';
+  result['embedder_version'] =
+      globals.cache.getStampFor(kTizenEmbedderStampName) ?? '';
```
JSUYA, line 662: This is the result from this code.
```yaml
  "dart_version": "3.6.0",
  "flutter_version": "3.27.1",
  "engine_version": "4d673e9827721f16b4b5f12d139b713feaa7ba3d",
  "embedder_version": "3b5d236f482e78e89b6b766c93b9777aba17788d"
```

I think this result is a bit more specific and each type can be distinguished.
like this. what do you think?
```yaml
  "dart_version": [
    "version" : "3.6.0",
    "hash" : "" # hash
  ],
  "flutter_version": [
    "version" : "3.27.1",
    "hash" : ""
  ],
  "engine_version": [
    "version" : "", # or "3.27.1"
    "hash" : "4d673e9827721f16b4b5f12d139b713feaa7ba3d"
  ],
  "embedder_version": [
    "version" : "",
    "hash" : "3b5d236f482e78e89b6b766c93b9777aba17788d"
  ]
```

lib/tizen_plugins.dart
```diff
@@ -629,3 +633,35 @@ Future<void> renderTemplateToFile(
...
+  FlutterProject project,
+) async {
+  final File extraInfoFile =
+      project.directory.childFile('.flutter-tizen-plugins-extra-info');
+
+  final List<TizenPlugin> plugins = await findTizenPlugins(project);
+  final List<Map<String, Object>> pluginInfo = <Map<String, Object>>[];
+  for (final TizenPlugin plugin in plugins) {
+    pluginInfo.add(<String, Object>{
+      'name': plugin.name,
```
JSUYA, line 648: Can you add version information for your plugin?

lib/tizen_plugins.dart
```diff
@@ -629,3 +633,35 @@ Future<void> renderTemplateToFile(
...
+  final Map<String, Object?> result = <String, Object>{};
+  result['info'] =
+      'This is a generated file; do not edit or check into version control.';
+  result['plugins'] = pluginInfo;
+  result['date_created'] = globals.systemClock.now().toString();
+  result['dart_version'] = globals.flutterVersion.dartSdkVersion;
+  result['flutter_version'] = globals.flutterVersion.frameworkVersion;
+  result['engine_version'] =
+      globals.cache.getStampFor(kTizenEngineStampName) ?? '';
+  result['embedder_version'] =
```
JSUYA, line 661: It looks better without `_version`.
'dart', 'flutter', 'engine', 'embedder', 'flutter-tizen' <- hash

lib/tizen_plugins.dart
```diff
@@ -629,3 +633,82 @@ Future<void> renderTemplateToFile(
   await file.create(recursive: true);
   await file.writeAsString(renderedTemplate);
 }
+
+/// See: [_writeFlutterPluginsList] in flutter_plugins.dart
+Future<void> _writeTizenPluginsExtraInfo(
```
JSUYA, line 638: how about `_writeAppDependencyInfo`??

lib/tizen_plugins.dart
```diff
@@ -629,3 +633,82 @@ Future<void> renderTemplateToFile(
...
+      'This is a generated file; do not edit or check into version control.';
+  result['plugins'] = pluginInfo;
+  final Map<String, Object> dart = <String, Object>{};
+  dart['version'] = globals.flutterVersion.dartSdkVersion;
+
+  final Map<String, Object> flutter = <String, Object>{};
+  flutter['version'] = globals.flutterVersion.frameworkVersion;
+
+  final Map<String, Object> flutterTizen = <String, Object>{};
+  flutterTizen['revision'] = globals.flutterVersion.frameworkRevisionShort;
```
JSUYA, line 693: I don't think we need to use short hash. See here. https://github.com/flutter-tizen/flutter-tizen/tree/master/bin/internal

lib/tizen_plugins.dart
```diff
@@ -629,3 +633,82 @@ Future<void> renderTemplateToFile(
...
+    }
+    pluginInfo.add(<String, Object>{
+      'name': plugin.name,
+      'version': version,
+    });
+  }
+
+  final Map<String, Object?> result = <String, Object>{};
+  result['info'] =
+      'This is a generated file; do not edit or check into version control.';
```
JSUYA, line 684: It looks like 'info' isn't needed.

lib/tizen_plugins.dart
```diff
@@ -629,3 +633,82 @@ Future<void> renderTemplateToFile(
...
+    }
+    final Object? packages = contents['packages'];
+    if (packages == null || packages is! YamlMap) {
+      return null;
+    }
+    return packages;
+  }
+
+  final File appDepsJson = project.directory.childFile('app.deps.json');
+
```
JSUYA, line 664: please remove empty space

JSUYA (approved):
Overall it's good. I left a few comments.

lib/tizen_plugins.dart
JSUYA: This file is created at packaging time and is created in the build directory. Therefore, there is no situation when the version-control program can detect it. so, I think it is unnecessary.

seungsoo47: Flutter makes plugins information files when running flutter commands.
When the Flutter-tizen collects plugins information, this file also is created, and located in flutter project by default. 
And then, it is copied to res folder during packaging time. 
So, I think it need to be remained.

JSUYA: um.. Apart from deleting the info text, I think the problem is that the app.deps.json file still remains in the working directory.
```
$ find -name "app.deps.json"
./app.deps.json
./build/tizen/tpk/tpkroot/res/flutter_assets/app.deps.json
./tizen/flutter/ephemeral/res/flutter_assets/app.deps.json
./tizen/bin/Release/tizen80/tpkroot/res/flutter_assets/app.deps.json
```
This file should only be included in the package file at packaging time, so there is no need to keep it in flutter_assets. And the file should also be deleted from the working root directory.

`./app.deps.json`
`./tizen/flutter/ephemeral/res/flutter_assets/app.deps.json`

seungsoo47: As we discussed offline, I have applied them.
- Change the file name from app.deps.json to .app.deps.json
- Move it to app_root/tizen/
- Add its name to app_root/tizen/.gitignore

Issue #610, opened 2025-04-09
Generated debug folder after release build


xiaowei-guan:
Reproduce steps:
1.flutter-tizen create sample003
2.flutter-tizen build tpk
3.check tizen/bin folder

```xml
guanxw@samsung10022:~/tizen/sample003$ flutter-tizen build tpk

Font asset "MaterialIcons-Regular.otf" was tree-shaken, reducing it from 1645184 to 1384 bytes (99.9% reduction). Tree-shaking can be disabled by providing the --no-tree-shake-icons flag when building your app.
The partner profile is used for signing.
Building a Tizen application in release mode...                    16.0s
✓ Built build/tizen/tpk/com.example.sample003-1.0.0.tpk (6.8MB)
guanxw@samsung10022:~/tizen/sample003$ ls
analysis_options.yaml  android  build  ios  lib  linux  macos  pubspec.lock  pubspec.yaml  README.md  sample003.iml  test  tizen  web  windows
guanxw@samsung10022:~/tizen/sample003$ cd tizen/
guanxw@samsung10022:~/tizen/sample003/tizen$ ls
App.cs  bin  flutter  obj  Runner.csproj  Runner.csproj.user  shared  tizen_dotnet_project.yaml  tizen-manifest.xml
guanxw@samsung10022:~/tizen/sample003/tizen$ tree
.
├── App.cs
├── bin
│   └── Debug
│       └── tizen80
│           ├── build.info
│           ├── com.example.sample003-1.0.0.tpk
│           ├── publish
│           ├── Runner.deps.json
│           ├── Runner.dll
│           ├── Runner.pdb
│           ├── Tizen.Flutter.Embedding.dll
│           ├── Tizen.Flutter.Embedding.pdb
│           ├── Tizen.Flutter.Embedding.xml
│           └── tpkroot
│               ├── author-signature.xml
│               ├── bin
│               │   ├── Runner.dll
│               │   ├── Runner.pdb
│               │   ├── Tizen.Flutter.Embedding.dll
│               │   ├── Tizen.Flutter.Embedding.pdb
│               │   └── Tizen.Flutter.Embedding.xml
│               ├── lib
│               │   ├── libapp.so
│               │   ├── libflutter_engine.so
│               │   └── libflutter_tizen.so
│               ├── res
│               │   ├── flutter_assets
│               │   │   ├── AssetManifest.bin
│               │   │   ├── AssetManifest.json
│               │   │   ├── FontManifest.json
[...]

PR #612, opened 2025-04-10
Fix `tz set` command failure

JSUYA:
Due to a bug in the `tz set` command, it fails to automatically generate tizen_dotnet_project.yaml.
So create it manually and call `tz set` command. This will be fixed in the next version of Tizen SDK.

https://github.com/flutter-tizen/flutter-tizen/issues/610

### DIFF
--- a/lib/build_targets/package.dart
+++ b/lib/build_targets/package.dart
@@ -178,6 +178,24 @@ class DotnetTpk extends TizenPackage {
         'Install the latest .NET SDK from: https://dotnet.microsoft.com/download',
       );
     }
+
+    // TODO(jsuya): Due to a bug in the `tz set` command, it fails
+    // to automatically generate tizen_dotnet_project.yaml. So
+    // create it manually and `tz set` command. This will be fixed
+    // in the next version of Tizen SDK.
+    // https://github.com/flutter-tizen/flutter-tizen/issues/610
+    if (!tizenProject.hostAppRoot
+        .childFile('tizen_dotnet_project.yaml')
+        .existsSync()) {
+      await _processUtils.run(
+        <String>[
+          'touch',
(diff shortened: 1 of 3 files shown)


lib/build_targets/package.dart
```diff
@@ -178,6 +178,24 @@ class DotnetTpk extends TizenPackage {
...
     }
+
+    // TODO(jsuya): Due to a bug in the `tz set` command, it fails
+    // to automatically generate tizen_dotnet_project.yaml. So
+    // create it manually and `tz set` command. This will be fixed
+    // in the next version of Tizen SDK.
+    // https://github.com/flutter-tizen/flutter-tizen/issues/610
+    if (!tizenProject.hostAppRoot
+        .childFile('tizen_dotnet_project.yaml')
+        .existsSync()) {
```
swift-kim, line 189: Simply `createSync(recursive: true)` without the `if` statement.

`touch` is not a valid command on Windows.

lib/tizen_sdk.dart
```diff
@@ -253,25 +253,27 @@ class TizenSdk {
...
         if (buildType != null) ...<String>['-b', buildType],
         if (signingProfile != null) ...<String>['-s', signingProfile],
-        '-w',
-        workingDirectory,
       ],
+      workingDirectory: workingDirectory,
     );
 
+    if (result.exitCode != 0) {
+      throwToolExit('Failed to build .NET application:\n$result');
```
swift-kim, line 268: The return type of this method is `RunResult`. A better practice would be to just return the `result` here.

lib/build_targets/package.dart
JSUYA: Thank you for your advice.

seungsoo47 (approved):
LGTM

PR #620, opened 2025-04-23
Add user-defined pixel ratio

JSUYA:
Add UserPixelRatio to set the pixel ratio. Users can adjust the base scale by setting this value before calling OnCreate().
If the value is not set or is less than 0, the pixel ratio is determined by the Tizen base scale calculation. https://docs.tizen.org/application/native/guides/ui/efl/multiple-screens/


JSUYA:
+) @myroot

seungsoo47 (approved):

PR #463, opened 2022-11-17
Update the known plugins list

swift-kim:
- Add `flutter_app_badger`, `flutter_secure_storage`, and `google_sign_in`.
- Remove `share_plus` (discontinued).
- Remove `video_player` and `webview_flutter` (the Tizen packages may not have the `_tizen` suffix).
  - `video_player_videohole`
  - `webview_flutter_lwe`

PR #622, opened 2025-05-08
Update the known plugins list

JSUYA:
- Add `flutter_webrtc`, `in_app_purchase`, `video_player` and `webview_flutter`.
- If there is more plugins besides *_tizen, use a different warning output format.

ex)
```
path_provider_tizen is available on pub.dev. Did you forget to add to pubspec.yaml?
shared_preferences_tizen is available on pub.dev. Did you forget to add to pubspec.yaml?
sqflite_tizen is available on pub.dev. Did you forget to add to pubspec.yaml?
[webview_flutter_tizen or webview_flutter_lwe] are available on pub.dev. Did you forget to add to pubspec.yaml?
```

### DIFF
--- a/lib/tizen_plugins.dart
+++ b/lib/tizen_plugins.dart
@@ -347,7 +347,7 @@ const Map<String, List<String>> _kKnownPlugins = <String, List<String>>{
   'shared_preferences': <String>[],
   'sqflite': <String>[],
   'url_launcher': <String>[],
-  'video_player': <String>['video_player_tizen', 'video_player_videohole', 'video_player_avplay'],
+  'video_player': <String>[],
   'wakelock_plus': <String>[],
   'webview_flutter': <String>['webview_flutter_tizen', 'webview_flutter_lwe'],
 };
--- a/lib/tizen_plugins.dart
+++ b/lib/tizen_plugins.dart
@@ -409,7 +409,10 @@ Future<void> _informAvailableTizenPlugins(FlutterProject project) async {
   final List<String> plugins = (await findPlugins(project)).map((Plugin p) => p.name).toList();
   for (final String plugin in plugins) {
     final String tizenPlugin = '${plugin}_tizen';
-    if (_kKnownPlugins.containsKey(plugin) && !plugins.contains(tizenPlugin)) {
+    if (_kKnownPlugins.containsKey(plugin) &&
+        !plugins.contains(tizenPlugin) &&
+        !_kKnownPlugins[plugin]!
+            .any((String recommendedPlugin) => plugins.contains(recommendedPlugin))) {
       final List<String> recommendedPlugins = _kKnownPlugins[plugin]!;
       if (recommendedPlugins.isEmpty) {
         globals.printWarning(
--- a/lib/tizen_plugins.dart
+++ b/lib/tizen_plugins.dart
@@ -323,30 +323,34 @@ Future<void> _generateEntrypointWithPluginRegistrant(
 }
 
 /// https://github.com/flutter-tizen/plugins
-const List<String> _kKnownPlugins = <String>[
-  'audioplayers',
-  'battery_plus',
-  'camera',
-  'connectivity_plus',
-  'device_info_plus',
-  'flutter_app_badger',
-  'flutter_secure_storage',
-  'flutter_tts',
-  'geolocator',
-  'google_maps_flutter',
-  'google_sign_in',
-  'image_picker',
-  'integration_test',
-  'network_info_plus',
-  'package_info_plus',
-  'path_provider',
-  'permission_handler',
-  'sensors_plus',
-  'shared_preferences',
-  'sqflite',
-  'url_launcher',
-  'wakelock',
-];
+const Map<String, List<String>> _kKnownPlugins = <String, List<String>>{
+  'audioplayers': <String>[],
+  'battery_plus': <String>[],
+  'camera': <String>[],
+  'connectivity_plus': <String>[],
+  'device_info_plus': <String>[],
+  'flutter_app_badger': <String>[],
+  'flutter_secure_storage': <String>[],
+  'flutter_tts': <String>[],
+  'flutter_webrtc': <String>[],
+  'geolocator': <String>[],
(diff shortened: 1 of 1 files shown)


lib/tizen_plugins.dart
```diff
@@ -405,9 +409,15 @@ Future<void> _informAvailableTizenPlugins(FlutterProject project) async {
   final List<String> plugins = (await findPlugins(project)).map((Plugin p) => p.name).toList();
   for (final String plugin in plugins) {
     final String tizenPlugin = '${plugin}_tizen';
-    if (_kKnownPlugins.contains(plugin) && !plugins.contains(tizenPlugin)) {
-      globals.printWarning(
-          '$tizenPlugin is available on pub.dev. Did you forget to add to pubspec.yaml?');
+    if (_kKnownPlugins.containsKey(plugin) && !plugins.contains(tizenPlugin)) {
```
swift-kim, line 412: - Note that `video_player_videohole` and `video_player_avplay` are standalone plugins and do not require the `video_player` dependency.
- A user who added both `video_player` and `video_player_avplay` as dependencies will still be warned since the condition only checks for `_tizen`.

My recommendation is not to add `video_player` and `webview_flutter` to avoid additional complexity.

JSUYA: As you said, `video_player_videohole` and `video_player_avplay` are standalone plugins, so this code is not correct now.
(I've already seen the deleted history [here](https://github.com/flutter-tizen/flutter-tizen/pull/463).)
However, if the user used `video_player`, I think our tool could recommend `video_player_tizen`. (webview_flutter case is also same.)

So I'm going to modify it like this.
```diff
-'video_player': <String>['video_player_tizen', 'video_player_videohole', 'video_player_avplay'],
+'video_player': <String>[],
```

When a flutter app uses video_player and webview_flutter, an exception occurs at runtime on tizen. I think it's little strange that these plugins do not output a message, unlike other plugins.

swift-kim: The second point in my above comment. A user who uses `video_player_avplay` or `video_player_lwe` will be warned repeatedly with irrelevant information. Exception is normal. Spamming wrong suggestions is not normal.

JSUYA: A warning from `video_player_tizen` is required when `video_player_avplay` and `video_player` are used together.
Because `video_player_avplay` is standalone and its use is not related to the use of `video_player`.

The `webview_flutter_lwe` is a bug. I have made an additional patch, can you please check it again?

swift-kim: `video_player` and `video_player_avplay` are technically unrelated but the presence of `video_player_avplay` implies that `video_player_tizen` is never used by the app. The tool should be conservative and should not print any message in the following case because Flutter is cross-platform.

```yaml
dependencies:
  video_player:  # not used by Tizen, used by other platforms
  video_player_avplay:  # only used by Tizen
```

lib/tizen_plugins.dart
```diff
@@ -405,9 +409,18 @@ Future<void> _informAvailableTizenPlugins(FlutterProject project) async {
...
+        !plugins.contains(tizenPlugin) &&
+        !_kKnownPlugins[plugin]!
+            .any((String recommendedPlugin) => plugins.contains(recommendedPlugin))) {
+      final List<String> recommendedPlugins = _kKnownPlugins[plugin]!;
+      if (recommendedPlugins.isEmpty) {
+        globals.printWarning(
+            '$tizenPlugin is available on pub.dev. Did you forget to add to pubspec.yaml?');
+      } else {
+        globals.printWarning(
+            '[${recommendedPlugins.join(', ')}] are available in pub.dev. Did you forget to add to pubspec.yaml?');
```
swift-kim, line 422:
```suggestion
            '[${recommendedPlugins.join(', ')}] are available on pub.dev. Did you forget to add to pubspec.yaml?');
```
The message ("Did you forget to add") is not very clear. Does it mean the user should add all the listed packages to `pubspec.yaml`?

lib/tizen_plugins.dart
JSUYA: When the user adds `video_player_avplay`, `video_player` should not exist.
Because.. `video_player` is not indicated in the `video_player_avplay` guide.
This is incorrect usage and has nothing to do with the recommended message. 

I think that the case where `video_player` and `video_player_avplay` are written together is not reason that is written to be used as a tizen-only plugin for `video_player`.

If `video_player` is used together, it should be assumed that both the platform player(`video_player`) and the player for drm player are used in the app.
Even if `video_player` is selected as `_avplay` or depending on the platform in the code, if you use the flutter-tizen tool, it is necessary to guide the plugin for `video_player_tizen`.

(Even if the developer uses `audioplayer` or `network_info_plus` depending on the platform in the code, the tizen-only plugin is still recommended.

```yaml
dependencies:
  network_info_plus:  # not used by Tizen, used by other platforms
```

```dart
if (isTizen)
   # skip
else
   # Use network_info_plus API
```
Is this spam message in case that output the network_info_plus_tizen warning? )

lib/tizen_plugins.dart
JSUYA: I udpated code for more clear message.
````
[webview_flutter_tizen or webview_flutter_lwe] are available on pub.dev. Did you forget to add to pubspec.yaml?
```

lib/tizen_plugins.dart
swift-kim: > When the user adds `video_player_avplay`, `video_player` should not exist.

You should never assume the user only targets the Tizen platform.

I think we just have different points of view. I'd choose to do nothing (do not provide potentially misleading information) if I am unsure. I think the network_info example is a bit different because there's no other candidate like video_player_videohole or video_player_avplay. This is just my opinion. You can do it your way. Please remember that Flutter is cross-platform. That's the only important thing to mention.

lib/tizen_plugins.dart
swift-kim: nit: are should be is

JSUYA: Thanks, done.

seungsoo47 (approved):

Issue #623, opened 2025-05-09
Add flutter app type


JSUYA:
We need an app type for Tizen applications that are created using Flutter app model.

Currently, the app types of `tizen-manifest.xml` that generated with the `flutter-tizen create` command are `type="dotnet"` and `type="capp"`.

This `type` is known as the point at which the app launcher selects a loader (dotnet loader, native loader) on the device(or emulator).
If the type is `“dotnet”`, it uses the dotnet loader, and if it's not `"dotnet"`, it uses the native loader.
So it takes a lot of time to change "dotnet" type to "flutter-dotnet" type.
But `“capp”` can be changed to `“flutter”` or `“flutter-capp”`, etc.

The type can be unified as `“flutter”`.
It can also be `“dotnet-flutter”` instead of `“flutter-dotnet”`.
The goal is to create a separate flutter type that is distinct from the existing `dotnet` type.

swift-kim:
This requires a change on the platform side, which is against the original Flutter concept. Android and other platforms do not have a dedicated app model for Flutter.

JSUYA:
> This requires a change on the platform side

Which Platform side are you talking about? We are discussing this with appfw and are reviewing it internally. And we're planning to update the loader on TVs as well. We need to add flutter information(? or type) to tizen-manifest.xml to apply the optimizations supported by the platform.

> which is against the original Flutter concept. Android and other platforms do not have a dedicated app model for Flutter.

The app model we're talking about is the same concept that Tizen uses: dotnet app, native app, and web app. This is just one way of categorizing the types we use within tizen.
The App type we are talking about here is different from the app type of Android or other platforms.
This can be utilized for statistics and management in TV store. It can also be used by loaders inside Tizen to optimize loaders for performance, etc.

um... I don't know why you are mentioning flutter concept for this point.
Does this mean that flutter don't make applications for only one platform?

If you don't understand this task, you can reach out to me directly. Or, if you want to talk about more, please explain in more detail.

PR #626, opened 2025-05-14
Add flutter_embedding metadata key

JSUYA:
This metadata key is used to identify apps written in flutter. It may be used for statistics in the store.
And in tizen runtime loader, loader can use this key to determine whether to optimize or not.

related issue: #623


JSUYA:
The inspiration for this work came from AndroidManifest.xml in android.
```
        <meta-data
            android:name="flutterEmbedding"
            android:value="2" />
```
Android is known to use this metadata to enable v2 embedding.
Tizen's embedding doesn't have a version, so I made it a bool type.
If you have a better idea for the key name or something, I'd appreciate it.

PR #627, opened 2025-05-14
Using precreated native window

JSUYA:
Tizen can provide a precreated window handle to an application running as a candidate process. Before creating the application's window, check whether it is a precreated window handle and pass it to the embedder. Currently, the candidate process on TV only applies to in-house applications and is not supported by the FlutterApplication class, which is based on CoreUIApplication.

This feature has been tested on RPI-based tizen profile(tizen9.0). Also, it only works if launched via `app_launcher -s`.


JSUYA:
+ https://github.com/flutter-tizen/embedder/pull/89

seungsoo47 (approved):
LGTM
Please add the 'window_handle' variable to flutter_app.h

JSUYA:
> Please add the 'window_handle' variable to flutter_app.h

Hi Could you explain more this? Why `flutter_app.h` need to `window_handle`?

PR #634, opened 2025-06-04
Update doc configure device

JSUYA:
The default version on the official site is Tizen 9.0.
So update URL of the image for rpi4 to the Tizen 9.0.

### DIFF
--- a/doc/configure-device.md
+++ b/doc/configure-device.md
@@ -85,9 +85,8 @@ You need a Linux PC (for flashing), a Raspberry Pi 4 board, and a micro SD card 
    ```
 
 1. [Download a Tizen OS image](https://docs.tizen.org/platform/developing/flashing-rpi/#download-binaries). Visit the links below and download two `.tar.gz` files.
-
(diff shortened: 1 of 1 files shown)


JSUYA:
> The [official guide](https://docs.tizen.org/platform/developing/flashing-rpi/) also mentions "Tizen" and "Tizen-unified" (along with "Tizen-9.0" and "Tizen-9.0-Unified"). It doesn't seem to be a good idea to update our document every time a new Tizen version is released.

I don't think it's necessary to write as much information as the official guide.
The reason for changing to 9.0 is that Tizen-Unified continues to generate the latest snapshots.
In addition, when a new version is released, users will download the latest version through the documentation in a situation where support for the new version is unclear.
I think a stable snapshot link is appropriate to avoid unnecessary problems.
I will changes when support for the new version stabilizes or new policies are decided.

I am well aware that it is good to reduce management points.
If there is something you want, please let me know specifically.

seungsoo47 (approved):

PR #638, opened 2025-06-25
Implement merged UI and platform thread

seungsoo47:


seungsoo47:
This pr should be re-verified after embedder pr is merged. 
See https://github.com/flutter-tizen/embedder/pull/104

PR #649, opened 2025-07-25
[ci] Add an action to automatically close long-inactived issues

JSUYA:
Close issue pages that have been inactive for more than 60 days.
but do not close issues that have a "bug", "enhancement" or "keep" label.

+) And add license information.

### DIFF
--- a/.github/workflows/analyze.yml
+++ b/.github/workflows/analyze.yml
@@ -1,3 +1,7 @@
+# Copyright 2025 The Flutter Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
 name: Analyze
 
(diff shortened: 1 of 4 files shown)


.github/workflows/analyze.yml
```diff
@@ -1,3 +1,7 @@
+# Copyright 2025 The Flutter Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
```
swift-kim, line 3: I authored these workflow files from scratch. I don't believe you need to credit the Flutter contributors.

JSUYA: It was my mistake. I've uploaded a patch. Thank you for review.

PR #660, opened 2025-08-25
Support pointing device setting

JSUYA:
If the application does not support pointing device(mouse) events, app developer can guide the D-PAD or display a related toast popup.
Tizen Web application development supports related settings.
https://developer.samsung.com/smarttv/develop/guides/user-interaction/mouse.html

Flutter-tizen will supports setting whether to support in host applications.
App developers can use `IsPointingDeviceSupport` and `IsFloatingMenuSupport`. Both default to true.

If `IsPointingDeviceSupport` is false, mouse input will not work and and the D-PAD will show with the floating menu.
When the user clicks the mouse, a toast pop-up notification is displayed only once (it is reset when the computer is rebooted).

If `IsFloatingMenuSupport` is false, pointing device input will not work and the floating menu will not be displayed.
(Even if `IsFloatingMenuSupport` is false and `IsPointingDeviceSupport` is true, pointing device input will not work.)

In App.cs
```cs
protected override void OnCreate()
{
    IsPointingDeviceSupport = false; // Set before calling base.OnCreate()
    base.OnCreate();

    GeneratedPluginRegistrant.RegisterPlugins(this);
}
```

related issue: https://github.com/flutter-tizen/embedder/issues/111
embedder: https://github.com/flutter-tizen/embedder/pull/112


embedding/cpp/include/flutter_app.h
```diff
@@ -111,9 +111,27 @@ class FlutterApp : public flutter::PluginRegistry {
 
   // The user-defined pixel ratio.
   //
-  //  Defaults to the device pixel ratio if the value is 0.
+  // Defaults to the device pixel ratio if the value is 0.
   double user_pixel_ratio_ = 0.0;
 
+  // Whether the app should be supported pointing device(mouse) or not.
```
seungsoo47:
```suggestion
  // Whether the app should support pointing devices (mouse).
```
Please check other comments, too.

embedding/cpp/include/flutter_app.h
```diff
@@ -111,9 +111,27 @@ class FlutterApp : public flutter::PluginRegistry {
...
+  //
+  // If false, pointing device input will not work and and the D-PAD will show
+  // with the floating menu. If true and is_floating_menu_support is false,
+  // pointing device input will not work.
+  //
+  // It only works on TV.
+  bool is_pointing_device_support = true;
+
+  // Whether the app should be supported floating menu or not.
```
seungsoo47:
```suggestion
  // Whether the app should support floating menus
```

embedding/csharp/Tizen.Flutter.Embedding/FlutterApplication.cs
```diff
@@ -86,6 +86,24 @@ public class FlutterApplication : CoreUIApplication, IPluginRegistry
...
         protected bool IsTopLevel { get; set; } = false;
 
+        /// <summary>
+        /// Whether the app should be supported pointing device(mouse) or not.
+        /// If false, pointing device input will not work and and the D-PAD will show with the floating menu.
+        /// If true and IsFloatingMenuSupport is false, pointing device input will not work.
+        /// 
+        /// It only works on TV.
+        /// </summary>
+        protected bool IsPointingDeviceSupport { get; set; } = true;
```
seungsoo47, line 96: I have a question: Is there a place where app developers can search for these APIs?

JSUYA: As far as I know, currently, there is no guide and API reference about FlutterApplication(cs, cpp). someday, we will need to make a guide for that APIs  .

embedding/cpp/include/flutter_app.h
JSUYA: done

embedding/cpp/include/flutter_app.h
JSUYA: done

JSUYA:
@xiaowei-guan
Could you please review this patch and guide CP to set `IsPointingDeviceSupport=false` or `IsFloatingMenuSupport=false`?

PR #662, opened 2025-08-29
Add the path to the user's libs to be copied.

JSUYA:
If ./gen/custom_def.prop exists in the native path, it will be parsed and added to `USER_LIB_PATH` as the path.

### DIFF
--- a/lib/build_targets/plugins.dart
+++ b/lib/build_targets/plugins.dart
@@ -189,6 +189,34 @@ class NativePlugins extends Target {
         );
       }
 
+      Map<String, String> parseCustomDefProp(File file) {
+        final Map<String, String> properties = <String, String>{};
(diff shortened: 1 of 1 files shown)


JSUYA:
> How do you think about providing build_def.prop template like the project_def.prop?

Could you explain that in a bit more detail?
build_def.prop and project_def.prop are features provided by the Tizen CLI.
https://docs.tizen.org/application/tizen-studio/native-tools/project-conversion/

When user SO files exist in the Native Plugin, only the SO files located in the lib, lib/{arch}, and lib/{arch}/{version} paths are packaged together.
This format is maintained in most cases.

I had expected to be able to use it when a separate LIBS was needed.
However, since it is not an official file, I will apply this patch after further consideration.

Issue #554, opened 2024-06-03
Support x64 emulator (Tizen 8.0)


JSUYA:
A 64-bit emulator was added in Tizen 8.0.
https://www.tizen.org/blogs/sungjinlee/2023/announcing-tizen-studio-5.5-release
Therefore, common profile + x86_64 arch builds should be supported.
`flutter-tizen build tpk --debug --target-arch x86_64 -p common`
To support this environment, the x86_64 artifact must be supported in both the engine and embedder.

swift-kim:
We can simply use the term "x64" (instead of x86_64) for consistency with the current codebase and flutter_tools.

PR #665, opened 2025-09-12
Support x64 emulator

JSUYA:
issue - https://github.com/flutter-tizen/flutter-tizen/issues/554
embedder - https://github.com/flutter-tizen/embedder/pull/118
engine - https://github.com/flutter-tizen/flutter/pull/13

- Supports debug, profile, and release builds for the x64 emulator.
(This build only works when the host is a Linux x64 environment.)
- To build an app for the x64 emulator, the API version must be 8.0 or higher.
(The emulator64 rootstrap in the Tizen SDK was released starting from version 8.0.)


lib/build_targets/utils.dart
```diff
@@ -30,18 +30,27 @@ String getBuildConfig(BuildMode buildMode) {
...
+    apiVersion = '8.0';
+  } else if (version >= Version(6, 5, 0)) {
     apiVersion = '6.5';
   } else {
     apiVersion = '6.0';
   }
+
+  if (arch == 'x86_64') {
+    arch = 'x64';
+  }
```
seungsoo47: I am not sure but it seems this code should be moved to line 42.

lib/tizen_artifacts.dart
```diff
@@ -42,6 +45,13 @@ class TizenArtifacts extends CachedArtifacts {
       final HostPlatform hostPlatform = getCurrentHostPlatform();
       assert(hostPlatform != HostPlatform.linux_arm64,
           'Artifact $artifact not available on Linux arm64.');
+      // if (arch == 'x64') {
+      //   return _cache
+      //       .getArtifactDirectory('engine')
+      //       .childDirectory(getNameForHostPlatform(hostPlatform))
+      //       .childFile('gen_snapshot')
+      //       .path;
+      // }
```
seungsoo47: If this code are not necessary, how about removing it?

lib/tizen_device.dart
```diff
@@ -119,7 +120,7 @@ class TizenDevice extends Device {
 
   @override
   Future<bool> supportsRuntimeMode(BuildMode buildMode) async {
-    if (_isLocalEmulator) {
+    if (getCapability('cpu_arch') == 'x86') {
```
seungsoo47, line 123:
```suggestion
    if (getCapability('cpu_arch') == 'x86' || getCapability('cpu_arch') == 'x86_64')) {
```
Don't you need 'x86_64', too?

lib/build_targets/utils.dart
JSUYA: done

lib/tizen_artifacts.dart
JSUYA: done

lib/tizen_device.dart
JSUYA: done

JSUYA: Sorry, it was my mistake. Since we supports the release and profile(runtimeMode) for x64, ‘x86_64’ should not be added. I will revert the code.

seungsoo47: Oh, I see. Thank you for explaining.

PR #689, opened 2025-11-17
In release mode, remove dev_dependencies plugins

seungsoo47:
Rework the code 'Remove dev dependencies from release build'(https://github.com/flutter-tizen/flutter-tizen/pull/621) which was reverted from https://github.com/flutter-tizen/flutter-tizen/pull/663.
- Not checking 'explicit-package-dependencies' flag
- Change computeExclusiveDevDependencies() to computeTransitiveDependencies()
- Fix related TestCases

For upstream code, see https://github.com/flutter/flutter/pull/161343/files#diff-68361f6fc9aafa5fa7de042eec4603dd4f321eb783a00edd57780ef78c1094bb

### DIFF
--- a/lib/build_targets/plugins.dart
+++ b/lib/build_targets/plugins.dart
@@ -62,8 +62,7 @@ class NativePlugins extends Target {
 
     // Check if there's anything to build.
     List<TizenPlugin> nativePlugins = await findTizenPlugins(project, cppOnly: true);
-    final bool releaseMode = buildInfo.buildInfo.mode.isRelease;
-    if (releaseMode) {
+    if (buildInfo.buildInfo.mode.isRelease) {
       nativePlugins = nativePlugins.where((TizenPlugin p) => !p.isDevDependency).toList();
     }
     if (nativePlugins.isEmpty) {
--- a/lib/tizen_plugins.dart
+++ b/lib/tizen_plugins.dart
@@ -384,8 +384,7 @@ Future<void> injectTizenPlugins(FlutterProject project) async {
   if (tizenProject.existsSync()) {
     List<TizenPlugin> cppPlugins = await findTizenPlugins(project, cppOnly: true);
     List<TizenPlugin> dotnetPlugins = await findTizenPlugins(project, dotnetOnly: true);
-    final bool releaseMode = FlutterCommand.current?.getBuildMode().isRelease ?? false;
-    if (releaseMode) {
+    if (FlutterCommand.current?.getBuildMode().isRelease ?? false) {
       cppPlugins = cppPlugins.where((TizenPlugin p) => !p.isDevDependency).toList();
       dotnetPlugins = dotnetPlugins.where((TizenPlugin p) => !p.isDevDependency).toList();
     }


lib/build_targets/plugins.dart
```diff
@@ -61,7 +61,11 @@ class NativePlugins extends Target {
     final tizenProject = TizenProject.fromFlutter(project);
 
     // Check if there's anything to build.
-    final List<TizenPlugin> nativePlugins = await findTizenPlugins(project, cppOnly: true);
+    List<TizenPlugin> nativePlugins = await findTizenPlugins(project, cppOnly: true);
+    final bool releaseMode = buildInfo.buildInfo.mode.isRelease;
+    if (releaseMode) {
```
JSUYA: `releaseMode` variable doesn't seem necessary. It's not reused.
```dart
if (buildInfo.buildInfo.mode.isRelease)
```

lib/tizen_plugins.dart
```diff
@@ -371,8 +382,13 @@ Future<void> _ensurePluginsReadyForTizenTooling(FlutterProject project) async {
 Future<void> injectTizenPlugins(FlutterProject project) async {
   final tizenProject = TizenProject.fromFlutter(project);
   if (tizenProject.existsSync()) {
-    final List<TizenPlugin> cppPlugins = await findTizenPlugins(project, cppOnly: true);
-    final List<TizenPlugin> dotnetPlugins = await findTizenPlugins(project, dotnetOnly: true);
+    List<TizenPlugin> cppPlugins = await findTizenPlugins(project, cppOnly: true);
+    List<TizenPlugin> dotnetPlugins = await findTizenPlugins(project, dotnetOnly: true);
+    final bool releaseMode = FlutterCommand.current?.getBuildMode().isRelease ?? false;
+    if (releaseMode) {
```
JSUYA: how about `if (FlutterCommand.current?.getBuildMode().isRelease ?? false)` ?

test/general/tizen_plugins_test.dart
```diff
@@ -52,14 +55,37 @@ flutter:
...
+      "name": "plugin_test",
+      "dependencies": ["some_dart_plugin"]
+    },
+    {
+      "name": "some_dart_plugin",
+      "dependencies": []
+    }
+  ],
+  "configVersion": 1
+}
```
JSUYA, line 77: flutter/packages/flutter_tools/test/general.shard/package_graph_test.dart
Upstream, it seems like JSON content is being written using jsonEncode. Do we need to update our test code as well?

Since the same file creation is repeated, can it be made into a function?
package_graph_test.dart
```
...
void writePackageGraph(List<Package> graph) {
...
```

lib/build_targets/plugins.dart
seungsoo47: Done

lib/tizen_plugins.dart
seungsoo47: Done

test/general/tizen_plugins_test.dart
seungsoo47: Thank you for the review.
Actually, I was considering creating the function.
But I would like to focus on this.
So, I'm going to create an issue related to the part you mentioned and implement the function later.

JSUYA: https://github.com/flutter-tizen/flutter-tizen/issues/695

JSUYA: +https://github.com/flutter-tizen/flutter-tizen/pull/705

PR #704, opened 2025-11-26
Fix to prevent selecting the wrong Tizen SDK path

JSUYA:
The existing code assumes the SDK path is found even if the actual directory doesn't exist(environment variable exists).
When selecting the SDK path, if the selected path is incorrect, choose the following method.


seungsoo47 (approved):
LGTM

PR #706, opened 2025-11-28
Change tizen-studio installer to tizen-sdk installer

seungsoo47:


.github/workflows/test.yml
```diff
@@ -51,7 +51,7 @@ jobs:
           sudo apt update
           sudo apt install -y libncurses5 python3.8 libpython3.8 gettext \
             libkf5itemmodels5 libkf5kiowidgets5 libkchart2 dbus-x11 gnome-keyring
-          curl https://download.tizen.org/sdk/Installer/tizen-studio_6.0/web-cli_Tizen_Studio_6.0_ubuntu-64.bin -o install.bin
+          curl https://download.tizen.org/sdk/Installer/tizen-sdk_10.0/web-cli_Tizen_SDK_10.0_ubuntu-64.bin -o install.bin
           chmod a+x install.bin
           ./install.bin --accept-license $HOME/tizen-studio
```
JSUYA, line 56: Is the path of env `TIZEN_SDK` set to `tizen-studio`?
I'm not sure what TizenSDK features will be more used here ci script in the future, but I'm a little concerned because the default installation path for TizenSDK isn't `tizen-studio`.

seungsoo47: Thank you for the review.
I'm not sure what you're looking for exactly, but I've verified that the new CLI installer installs the relevant libraries into the tizen-studio directory.
If you tell me more details, I will check it again. :)

Issue #717, opened 2025-12-10
Invalid guide text for the error message that rootstrap cannot be found.


JSUYA:
In install the SDK with VSCode + Tizen Extension case, package-manager-cli.bin(package-manager-cli.exe) is no longer provided.

[Guide message]
```
/home/{USER}/.tizen-extension-platform/server/sdktools/data/package-manager/package-manager-cli.bin install NativeCLI NativeToolchain-Gcc-9.2 IOT-Headed-6.0-NativeAppDevelopment
```

The guide message should change depending on the type of SDK installed.

PR #722, opened 2025-12-16
Add TizenSdkType to TizenSdk

JSUYA:
When installing the Tizen SDK using the VS Code Tizen extension,
the SDK is installed in the `.tizen-extension-platform` directory.
If installed via the extension, SDK-related errors should be displayed differently.
Therefore, if the SDK type is extension SDK, different guidance is shown when a package is missing.

```
exception:Error: The rootstrap iot-headed-7.0-device64.core could not be found.
To install missing package(s) :
Open the "Tizen: Package Manager" page in VS Code and install Tizen SDK 7.0 version.
```

https://github.com/flutter-tizen/flutter-tizen/issues/717

### DIFF
--- a/test/general/tizen_doctor_test.dart
+++ b/test/general/tizen_doctor_test.dart
@@ -122,7 +122,7 @@ class _FakeTizenSdk extends TizenSdk {
           logger: BufferLogger.test(),
           platform: FakePlatform(),
           processManager: FakeProcessManager.any(),
-          sdkType: TizenSdkType.studio,
+          sdkType: TizenSdkType.cli,
         );
 
   @override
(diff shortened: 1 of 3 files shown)


lib/tizen_sdk.dart
```diff
@@ -18,12 +18,22 @@ TizenSdk? get tizenSdk => context.get<TizenSdk>();
 
 File? get dotnetCli => globals.os.which('dotnet');
 
+/// The type of Tizen SDK installation.
+enum TizenSdkType {
+  /// VS code Tizen extension based SDK.
+  extension,
+
+  /// Tizen-Studio based SDK.
```
seungsoo47: As you know, Tizen-studio will soon be deprecated. Therefore, it seems necessary to use the Tizen CLI instead. Otherwise, there also is a way to maintain 3 options (vscode extension, tizen cli, tizen-studio).
Or does 'Tizen-Studio' refer to a directory path?

JSUYA: Changed `studio` to `cli`. cli is more appropriate.

seungsoo47 (approved):

Issue #721, opened 2025-12-12
Invalid SDK name info for  in `flutter-tizen doctor -v`


JSUYA:
```
[✓] Tizen toolchain - develop for Tizen devices [319ms]
    • Framework revision 5de16a151b (2 days ago), 2025-12-10 18:43:41 +0900
    • Engine revision b8ddcb37ac
    • Embedder revision 6cb5dadb53
    • Tizen Studio 10.0 at /home/junsu/.tizen-extension-platform/server/sdktools/data
    • .NET SDK 9.0.304 at /home/junsu/.tizen-extension-platform/server/sdktools/dotnet/dotnet
```

If tools are referencing Tizen SDK Extension, it should see Tizen SDK Extension 10.1.0 instead of Tizen Studio 10.0.

[before]
```
    • Tizen Studio 10.0 at /home/junsu/.tizen-extension-platform/server/sdktools/data
```

[after]
```
    • Tizen SDK Extension 10.1.0 at /home/junsu/.tizen-extension-platform/server/sdktools/data
```

PR #727, opened 2025-12-26
Update messages provided for missing packages

seungsoo47:
Modify messages written based on Tizen Studio to fit the VSCode Extension.

+) https://github.com/flutter-tizen/flutter-tizen/issues/721

### DIFF
--- a/lib/tizen_doctor.dart
+++ b/lib/tizen_doctor.dart
@@ -73,9 +73,6 @@ class TizenValidator extends DoctorValidator {
         .existsSync()) {
       missingPackages.add('NativeToolchain-Gcc-$gccVersion');
     }
-    if (!_tizenSdk.tzCli.existsSync()) {
-      missingPackages.add('NativeCLI');
-    }
     if (missingPackages.isNotEmpty && _tizenSdk.sdkType == TizenSdkType.extension) {
       messages.add(
           ValidationMessage.error('To install missing package(s) : ${missingPackages.join(' ')}\n'


lib/tizen_doctor.dart
```diff
@@ -101,8 +101,13 @@ class TizenValidator extends DoctorValidator {
...
 
     if (missingPackages.isNotEmpty) {
-      messages.add(ValidationMessage.error('To install missing packages, run:\n'
-          '$packageManager install ${missingPackages.join(' ')}'));
+      if (_tizenSdk.sdkType == TizenSdkType.cli) {
+        messages.add(ValidationMessage.error('To install missing packages, run:\n'
+            '$packageManager install ${missingPackages.join(' ')}'));
+      } else {
+        messages.add(const ValidationMessage.error('To install missing package(s) :\n'
+            'Open the "Tizen: Package Manager" page in VS Code and install Tizen SDK "6.0 IOT-Headed + Mobile".'));
```
JSUYA, line 117: The missingPackages list output has disappeared. It would be better to indicate which packages should be installed in each missingPackages.add() call on lines 72, 93, and 100.

lib/tizen_doctor.dart
```diff
@@ -154,7 +159,7 @@ class TizenValidator extends DoctorValidator {
     } else {
       final versionText = sdkVersion != null ? ' $sdkVersion' : '';
       messages.add(ValidationMessage(
-        'Tizen Studio$versionText at ${_tizenSdk.directory.path}',
+        'Tizen SDK$versionText at ${_tizenSdk.directory.path}',
```
JSUYA, line 170: I think it would be better to output differently depending on the SDK type.

lib/tizen_doctor.dart
seungsoo47: Could you explain in more detail?
I thought that if a developer installed [6.0 IOT-Headed + Mobile], all the missing packages would be installed.
Please let me know If I missed anything.

JSUYA: This line is that prints the list of all packages added to missingPackages. It is set to print if the Tizen CLI or native toolchain is not exist. (72,93 line)
These are packages that are installed by default when installing the extension in the Tizen Extension. Therefore, they are not related to 6.0 IoT Headed and I think they should be explained in a different way.

seungsoo47: Done

lib/tizen_doctor.dart
```diff
@@ -68,6 +68,19 @@ class TizenValidator extends DoctorValidator {
...
+        .existsSync()) {
+      missingPackages.add('NativeToolchain-Gcc-$gccVersion');
+      if (_tizenSdk.sdkType == TizenSdkType.extension) {
+        messages.add(
+            ValidationMessage.error('To install missing package(s) : ${missingPackages.first}\n'
+                'Delete the Tizen home directory and reinstall the VSCode Tizen extension.\n'
+                'Note: Before proceeding, back up any important data (ex. certificate files).'));
+        return false;
+      }
+    }
```
JSUYA, line 82: I don't understand why this if condition moved. and I think, 'Tizen Home directory' is an incorrect expression. It would be more appropriate to say `Tizen SDK directory` or the path of the extension. Furthermore, the same warning should be shown not only for gcc but also for the tz command (if the tz command is a default installation target).

seungsoo47: Thank you for the review.

The existing code checks for the presence of `tizen` instead of `tz`, so I have added the necessary code to check for 'tz'. I have also confirmed that the `gcc toolchain` and `tz` are installed by default. Due to this, I have moved the if condition. Additionally, when reinstalling the extension, deleting the Tizen SDK directory (`~/.tizen-extension-platform/server/sdktools/data`) does not have any effect. Instead, we should delete the Tizen Home directory (`~/.tizen-extension-platform`), not the Tizen SDK directory.

Please let me know if I missed something.

lib/tizen_doctor.dart
```diff
@@ -68,6 +68,22 @@ class TizenValidator extends DoctorValidator {
...
     var result = true;
 
+    if (!_tizenSdk.toolsDirectory
+        .childDirectory('arm-linux-gnueabi-gcc-$gccVersion')
+        .existsSync()) {
+      missingPackages.add('NativeToolchain-Gcc-$gccVersion');
+    }
+    if (!_tizenSdk.tzCli.existsSync()) {
+      missingPackages.add('NativeCLI');
+    }
```
JSUYA: It seems that this line was added because I mentioned the tz command. This is basically very similar to the tizenCli check on line 87. For now, it would be better to delete this line.

Issue #696, opened 2025-11-20
Support Tizen SDK 10.0


JSUYA:
With the TizenStudio SDK update from  `TizenStudio SDK` 6.1 to  `TizenSDK` 10.0, TizenStudio is no longer supported. (`TizenStudio SDK` -> `TizenSDK`)
https://developer.tizen.org/development/tizen-studio/tizen-studio-deprecation-notice/

Currently, you can still install Tizen Studio 6.1 and flutter-tizen will work on it, but if user install the latest SDK, flutter-tizen will not work.

TizenSDK 10.0 provides functionality based on the VS Code extension. (While the CLI is still available), the Java-based manager GUI application (package, certificate, emulator) is no longer supported.
https://samsungtizenos.com/tools-download/
Therefore, the content and images in related guides (e.g., environment settings) must be updated.

Furthermore, the installation paths for the tz and tizen commands have changed, so we must refer to the new paths for TizenSDK 10.0.
https://github.com/flutter-tizen/flutter-tizen/issues/693

Of course, both TizenStudio SDK 6.1 and Tizen SDK 10.0 must be supported.

JSUYA:
+ https://github.com/flutter-tizen/flutter-tizen/pull/704

JSUYA:
https://github.com/flutter-tizen/flutter-tizen/pull/718

PR #732, opened 2026-01-28
Add Tizen Extension version detection and display

JSUYA:
- Add extensionVersion getter to TizenSdk class to read version from server-info.json
- Display Tizen Extension version in flutter doctor -v output

https://github.com/flutter-tizen/flutter-tizen/issues/721
https://github.com/flutter-tizen/flutter-tizen/issues/696

### DIFF
--- a/lib/tizen_sdk.dart
+++ b/lib/tizen_sdk.dart
@@ -174,11 +174,10 @@ class TizenSdk {
       if (decoded is Map<String, dynamic>) {
         return decoded['version'] as String?;
       }
-      return null;
     } on FormatException catch (error) {
       globals.printError('Failed to parse ${serverInfoFile.path}:\n$error');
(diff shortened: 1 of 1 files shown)


lib/tizen_sdk.dart
```diff
@@ -157,6 +159,28 @@ class TizenSdk {
...
+      return null;
+    }
+
+    try {
+      final String info = serverInfoFile.readAsStringSync();
+      final dynamic decoded = jsonDecode(info);
+      if (decoded is Map<String, dynamic>) {
+        return decoded['version'] as String?;
+      }
+      return null;
```
seungsoo47: How about moving this line to 182 line?

JSUYA: done

PR #739, opened 2026-02-05
Exclude dev_dependency dart plugins from generated registrant

JSUYA:
Do not include dev_dependency dart plugins when generating tizen/flutter/generated_main.dart.

https://github.com/flutter/flutter/pull/161828

### DIFF
--- a/test/general/tizen_plugins_test.dart
+++ b/test/general/tizen_plugins_test.dart
@@ -9,6 +9,7 @@ import 'package:file/memory.dart';
 import 'package:file_testing/file_testing.dart';
 import 'package:flutter_tizen/tizen_plugins.dart';
 import 'package:flutter_tools/src/base/file_system.dart';
+import 'package:flutter_tools/src/build_info.dart';
 import 'package:flutter_tools/src/cache.dart';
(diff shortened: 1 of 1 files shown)


lib/tizen_plugins.dart
```diff
@@ -307,7 +307,8 @@ Future<void> generateEntrypointWithPluginRegistrant(
   );
   final Uri mainUri = packageConfig.toPackageUri(mainFileUri) ?? mainFileUri;
   final List<String> dartEntrypoints = _findDartEntrypoints(mainFile);
-  final List<TizenPlugin> dartPlugins = await findTizenPlugins(project, dartOnly: true);
+  final List<TizenPlugin> dartPlugins =
+      (await findTizenPlugins(project, dartOnly: true)).where((p) => !p.isDevDependency).toList();
```
seungsoo47: I know that dev_dependency is only checked when the build is in release mode.
Please refer to https://github.com/camsim99/flutter/blob/master/packages/flutter_tools/lib/src/flutter_plugins.dart#L1259-L1264.

JSUYA: I added code to check for Release mode.

PR #743, opened 2026-02-10
Use Git wrapper for flutter command

JSUYA:
Replace ad-hoc git invocation with globals.gitc to match upstream behavior and improve Windows compatibility.

related pr : https://github.com/flutter/flutter/pull/172495

### DIFF
--- a/lib/commands/create.dart
+++ b/lib/commands/create.dart
@@ -4,6 +4,7 @@
 
 import 'package:flutter_tools/src/base/common.dart';
 import 'package:flutter_tools/src/base/file_system.dart';
+import 'package:flutter_tools/src/base/process.dart';
 import 'package:flutter_tools/src/base/terminal.dart';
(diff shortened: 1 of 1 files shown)


lib/commands/create.dart
```diff
@@ -396,30 +394,21 @@ class TizenCreateCommand extends CreateCommand {
...
+      <String>['checkout', '--', '.'],
       workingDirectory: directory.path,
     );
-    if (result.exitCode != 0) {
-      throwToolExit('Failed to run git checkout: ${result.stderr}');
-    }
-    result = globals.processManager.runSync(
-      <String>['git', 'clean', '-df', '.'],
+    globals.git.runSync(
+      <String>['clean', '-df', '.'],
```
seungsoo47: Could you explain why you removed the processing of the results?

JSUYA: my mistake. i have updated code

PR #746, opened 2026-02-13
Prevent blocking in Inspector pub root registration

JSUYA:
Add timeout to findExtensionIsolate call and use unawaited to avoid blocking the device log reader. This ensures the app doesn't freeze when DevTools Inspector fails to respond or times out.

- Add 3-second timeout to findExtensionIsolate VM service call
- Wrap _setupInspectorPubRootRegistration with unawaited to avoid awaiting the Future

### DIFF
--- a/lib/forwarding_log_reader.dart
+++ b/lib/forwarding_log_reader.dart
@@ -57,6 +57,8 @@ class ForwardingLogReader extends DeviceLogReader {
   static const _inspectorPubRootDirectoriesExtensionName =
       'ext.flutter.inspector.addPubRootDirectories';
 
+  static const _kInspectorFindIsolateTimeout = Duration(seconds: 3);
+
   // For Inspector pub root registration on hot restart
   StreamSubscription<vm_service.Event>? _isolateEventSubscription;
 
@@ -221,7 +223,7 @@ class ForwardingLogReader extends DeviceLogReader {
     try {
       final String? maybeIsolateId = (await vmService
               .findExtensionIsolate(_inspectorPubRootDirectoriesExtensionName)
-              .timeout(const Duration(seconds: 3)))
+              .timeout(_kInspectorFindIsolateTimeout))
           .id;
 
       if (maybeIsolateId case final isolateId?) {


JSUYA:
Without this PR, integration_test in plugins will not work.

lib/forwarding_log_reader.dart
```diff
@@ -209,7 +211,7 @@ class ForwardingLogReader extends DeviceLogReader {
     // tizen/flutter/generated_main.dart, which causes DevTools to incorrectly
     // set the pub root to tizen/flutter/ instead of the actual project root.
     // Without this fix, DevTools Inspector shows an empty widget tree.
-    await _setupInspectorPubRootRegistration(connectedVmService);
+    unawaited(_setupInspectorPubRootRegistration(connectedVmService));
```
seungsoo47, line 214: If you use the timeout() method, unawaited() probably won't be needed.
What do you think?

JSUYA: As you said, having just one of the two solutions is sufficient for functionality.

However, in my view, `setupInspectorPubRootRegistration()` method itself is logic that could be handled asynchronously without issue from the start. But if only `unwaited` is present, it will continue waiting for `findExtensionIsolate()` in another isolate.

The `timeout` depends on finding the `“ext.flutter.inspector.addPubRootDirectories”` extension. Therefore, the integration_test always incurs an additional 3 seconds of execution time.

Therefore, both pieces of logic must be included.

seungsoo47: I am sorry, but I don't understand what you said.

The findExtensionIsolate() function always times out when running profile mode or integration tests.
Even running profile mode and integration tests don't require finding the “ext.flutter.inspector.addPubRootDirectories”.

So how about handling the _setupInspectorPubRootRegistration() function only in debug mode?

JSUYA: As far as I know, integration_test using Flutter's drive defaults to debug mode. Therefore, we can't use build mode to determine whether to waiting for "ext.flutter.inspector.addPubRootDirectories."

seungsoo47: Thank you for explaining.

PR #758, opened 2026-04-10
Fallback to PATH for Linux native asset linker

JSUYA:
Use the upstream Linux compiler configuration when available, then fall back to PATH for ld.lld or ld while keeping compiler and archiver selection tied to the clang toolchain directory.


JSUYA:
@gemini review

PR #774, opened 2026-06-18
Change UI thread policy to kDefault

seungsoo47:
- Update C++ embedding to use kDefault instead of kRunOnSeparateThread
- Update C# embedding to match C++ threading behavior


JSUYA (approved):

JSUYA:
https://github.com/flutter-tizen/embedder/pull/171

PR #776, opened 2026-06-19
Skip pub upgrade when dependency inputs are unchanged

JSUYA:
`pub upgrade` ran on every flutter-tizen revision, even for Dart-only changes — re-resolving dependencies each time although the existing pubspec.lock was still valid.
Now run pub only when pubspec.lock or dart_tool/package_config.json is missing, or pubspec.yaml changed since the last resolution.

### DIFF
--- a/bin/internal/shared.bat
+++ b/bin/internal/shared.bat
@@ -52,6 +52,10 @@ GOTO :EOF
 
         REM Invalidate the cache.
         IF EXIST "%cache_dir%" RMDIR /S /Q "%cache_dir%"
+
+        REM Force pub to run again because the dependencies of flutter_tools
+        REM (a path dependency of flutter-tizen) may have changed.
+        IF EXIST "%ROOT_DIR%\pubspec.lock" DEL /F /Q "%ROOT_DIR%\pubspec.lock"
       )
 
       FOR /f %%r IN ('git rev-parse HEAD') DO SET revision=%%r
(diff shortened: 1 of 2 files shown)


bin/internal/shared.bat
```diff
@@ -100,12 +104,29 @@ GOTO :EOF
...
   :do_update_snapshot
     PUSHD "%ROOT_DIR%"
+      REM Run pub only when the dependency inputs changed. A revision change
+      REM that only touches Dart sources reuses the existing package
+      REM resolution.
+      IF NOT EXIST "%ROOT_DIR%\pubspec.lock" GOTO do_pub_upgrade
+      IF NOT EXIST "%ROOT_DIR%\.dart_tool\package_config.json" GOTO do_pub_upgrade
+      FOR /F %%i IN ('DIR /B /O:D "%ROOT_DIR%\pubspec.yaml" "%ROOT_DIR%\pubspec.lock"') DO SET pub_newer_file=%%i
+      FOR %%i IN (%ROOT_DIR%\pubspec.yaml) DO SET pub_yaml_timestamp=%%~ti
+      FOR %%i IN (%ROOT_DIR%\pubspec.lock) DO SET pub_lock_timestamp=%%~ti
```
seungsoo47:
```suggestion
      FOR %%i IN ("%ROOT_DIR%\pubspec.yaml") DO SET pub_yaml_timestamp=%%~ti
      FOR %%i IN ("%ROOT_DIR%\pubspec.lock") DO SET pub_lock_timestamp=%%~ti
```
If the workspace directory path contains spaces (e.g.  C:\My Projects\flutter-tizen ), the  FOR  loops will fail with syntax errors. How about wrapping the paths in double quotes?

seungsoo47 (commented):
a minor comment

bin/internal/shared.bat
JSUYA: done

seungsoo47: It seems you have changed other code.

JSUYA: I added more commit.

seungsoo47 (approved):

PR #781, opened 2026-07-02
Rename `.app.deps.json` to `app.deps.json` so it ships in the TPK

JSUYA:
The Tizen packager drops dot-prefixed files, so the hidden .app.deps.json was generated and copied into res/flutter_assets but never made it into the final TPK. Drop the leading dot so it is packaged like the other flutter_assets files.


JSUYA:
invalid - .app.deps.json is being packaged successfully.

Issue #784, opened 2026-07-06
Change UI thread policy to kDefault causes release TPKs to hang and crash on Samsung TV hardware


thachnb-harman:
**Flutter-Tizen version**
```
flutter-tizen 3.44.1 • channel master • https://github.com/flutter-tizen/flutter-tizen
Flutter 3.27.4 • revision b7401b5a7c (2025-02-03)
Engine • revision 6f3b557f2f
Dart 3.6.2 • DevTools 2.40.2
```

**Target device**

Samsung QN55QN80F (retail, Tizen 6.8, aarch64)

`sdb dlog` is blocked on retail devices; debug output captured via `flutter-tizen run --debug`.

**Describe the issue**

Commit `0de0113` ("Change UI thread policy to kDefault", PR #774) causes release TPKs to
hang and crash on Samsung TV hardware. After launch, our app hang and watchdog fires a SIGABRT (`DotNET onSigabrt`) and kills the app

The regression is bisect-confirmed: `899e49f` (the commit immediately before `0de0113`) works correctly; `0de0113` does not.

**Root cause:** `kDefault` merges the Flutter UI isolate onto the platform thread. On TV SoCs (slower cores, tighter watchdog) the merged thread becomes a bottleneck — rendering, platform-channel calls, and stream reads all compete on one thread. The watchdog kills the process when the main loop is unresponsive for too long.

**To reproduce**

1. Check out flutter-tizen at `0de0113`.
2. Build a release TPK with any Flutter app that loads images and uses platform channels.
3. Install on a Samsung TV (Tizen 6.0, aarch64).
4. Launch — images will not render; app dies after ~30–60 s.

**Log message**

Captured via `flutter-tizen run --debug` on a QN55QN80F:

```
[E] DotNET onSigabrt called on com.*** / DN_m.***(5786)
[E] onSigabrt called
```

**Bisection table**

| Commit | Description | Result |
|---|---|---|
| `36c8da0` | 3.44.1 tag | ✅ Images render, audio plays |
| `899e49f` | Embedder bump `972cc1c` | ✅ Images render, audio plays |
| `0de0113` | PR #774 — kDefault policy | ❌ Blank images → watchdog SIGABRT |

**Screenshots**

N/A — blank screen until watchdog kill.

**Code sample**

No minimal repro needed — any release TPK that loads remote images and uses platform channels
reproduces the crash on Samsung TV hardware.

**Proposed fix**

[...]

PR #785, opened 2026-07-08
Expose ui_thread_policy on FlutterEngine/FlutterApplication/FlutterApp

seungsoo47:
Since #774, the UI isolate's thread policy has been hardcoded to kDefault (merged onto the platform thread). On Tizen, the platform thread is also the app framework's main loop thread, so a long-blocking synchronous call on the UI isolate (e.g. via FFI) can leave the app unable to respond to app control/resume requests and get killed by the platform watchdog (see #784).

Add an optional ui_thread_policy parameter/property to FlutterEngine (C++ and C#), FlutterApp, and FlutterServiceApp, defaulting to kDefault to preserve existing behavior. Apps that need kRunOnSeparateThread can now opt into it, though the app remains responsible for making sure its Dart code never blocks indefinitely regardless of the policy chosen.

### DIFF
--- a/embedding/cpp/flutter_app.cc
+++ b/embedding/cpp/flutter_app.cc
@@ -11,7 +11,7 @@
 bool FlutterApp::OnCreate() {
   TizenLog::Debug("Launching a Flutter application...");
 
-  engine_ = FlutterEngine::Create(dart_entrypoint_);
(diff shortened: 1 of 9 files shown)


JSUYA (commented):
Please check FlutterServiceApplication too.

seungsoo47:
> Please check FlutterServiceApplication too.

Thanks. Done.

JSUYA (approved):

PR #810, opened 2026-08-21
Support version management commands (`channel`, `upgrade`)

JSUYA:
Adds version management to command(`channel`, `upgrade`), replacing manual `git checkout` of release tags.

  - `flutter-tizen channel` — lists release tags (newest tagged commit first), marking the current one.
  - `flutter-tizen channel <tag>` — checks out that tag (1:1 with `git tag`, e.g. `3.44.4-tizen.1.0.0` or legacy `3.44.4`), then re-runs
  bootstrap, `precache --force`, and `doctor`.
  - `flutter-tizen upgrade` — moves to the newest release tag. `--verify-only` only reports.

### DIFF
--- a/lib/commands/channel.dart
+++ b/lib/commands/channel.dart
@@ -1,0 +1,501 @@
+// Copyright 2026 Samsung Electronics Co., Ltd. All rights reserved.
+// Copyright 2014 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter_tools/src/base/common.dart';
+import 'package:flutter_tools/src/base/file_system.dart';
+import 'package:flutter_tools/src/base/io.dart';
+import 'package:flutter_tools/src/base/platform.dart';
+import 'package:flutter_tools/src/base/process.dart';
+import 'package:flutter_tools/src/cache.dart';
+import 'package:flutter_tools/src/commands/channel.dart';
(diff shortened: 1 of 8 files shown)


lib/commands/channel.dart
```diff
@@ -0,0 +1,496 @@
...
+    globals.printStatus('Switching flutter-tizen to ${release.tag}...');
+    await checkoutDetached(_gitWrapper, repoRoot, commit, force: boolArg('force'));
+    printRecovery(repoRoot: repoRoot, branch: branch, headCommit: head);
+    await runPostSwitchSetup(
+      repoRoot: repoRoot,
+      targetTag: release.tag,
+      cacheArtifacts: boolArg('cache-artifacts'),
+      previousFlutterVersion: pinnedFlutterVersion,
+      runLauncher: _runLauncher,
+    );
```
seungsoo47, line 271: channel.dart calls runPostSwitchSetup() directly without suppressing the welcome-message flag. upgrade.dart wraps the same call with setShouldRedisplayWelcomeMessage(false) / finally { ...(true) } to stop the "Welcome to Flutter!" banner from popping up during the automated subprocess chain. Switching versions via flutter-tizen channel <tag> can unexpectedly print that banner.
How about adding the setShouldRedisplayWelcomeMessage(false) / setShouldRedisplayWelcomeMessage(true)?

JSUYA: done

seungsoo47 (approved):

